PR #70, opened 2015-03-17
Migrate `next_*` methods to return `Option`

frewsxcv:
In the Rust standard library, the `next` method on the `Iterator` trait
returns an `Option`. Prior to this commit, the `next` and `next_*`
methods return `Err<_, ()>`. In an attempt to make some of the code more
idiomatic, I migrated these methods to return `Option` instead.

### DIFF
--- a/src/color.rs
+++ b/src/color.rs
@@ -72,10 +72,10 @@ impl Color {
     ///
     /// FIXME(#2) Deprecated CSS2 System Colors are not supported yet.
     pub fn parse(input: &mut Parser) -> Result<Color, ()> {
-        match try!(input.next()) {
-            Token::Hash(value) | Token::IDHash(value) => parse_color_hash(&*value),
-            Token::Ident(value) => parse_color_keyword(&*value),
-            Token::Function(name) => {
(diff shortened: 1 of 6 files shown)


SimonSapin:
Thanks for the good intentions, but using `Result<_, ()>` instead of `Option<_>` was very deliberate. It enables using the `try!` macro, which I do extensively in Servo.

I’ve experimented with [a generalized `try!` macro](https://github.com/SimonSapin/rust-std-candidates/blob/master/triable/lib.rs) that works on `Option` as well as `Result`, but I’m not very happy with it since it doesn’t convert between them.

Until returning `Option` can be as ergonomic, I prefer sticking with `Result`.

frewsxcv:
> since it doesn’t convert between them

What do you mean by this?

SimonSapin:
I mean that you can use this modified `try!` macro with an `Option<T>` expression in a context where the return type is `Option<U>`, or with a `Result<T, E>` expression where it is `Result<U, E>`, but not e.g. with a `Option<T>` expression where the return type is `Result<U, E>`.

PR #73, opened 2015-04-02
Clamp to MAX / INFINITY rather than panic on overflow in number parsing

SimonSapin:
r? @jdm

This also changes the numeric types in tokens from `f64` / `i64` to `f32` / `i32`, which might requires some changes in Servo.

### DIFF
--- a/src/color.rs
+++ b/src/color.rs
@@ -309,18 +309,18 @@ fn parse_color_function(name: &str, arguments: &mut Parser) -> Result<Color, ()>
         // Either integers or percentages, but all the same type.
         match try!(arguments.next()) {
             Token::Number(ref v) if v.int_value.is_some() => {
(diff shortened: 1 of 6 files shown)


jdm:
I haven't looked at the merge commit; the rest look fine to me. Should I be looking at the merge?

SimonSapin:
The merge includes some fairly simple conflict resolutions, as well as a replacement of `Float::powf` with `f64::powf` (corresponding to a Rust upgrade).

By the way, the commit to use in Servo (unless Servo does another Rust upgrade) is the one before the merge.

jdm:
I don't think there's a burning need to adopt these changes before the next upgrade. Looks good to me!

SimonSapin:
> Looks good to me!

I’ll take that as r+ :)

PR #77, opened 2015-07-13
Allow colors to be serialized with `serde`.

pcwalton:
`serde` is an optional Cargo feature.

r? @SimonSapin


SimonSapin:
Looks like homu is not set up here. CC @metajack

PR #83, opened 2015-08-04
clamp rgb values to [0,1]

### LINKED ISSUE #76: RGB values need clamping
Per http://www.w3.org/TR/CSS2/syndata.html#value-def-color, integers and percentages above the maximum should be clamped.

This breaks tests/wpt/css-tests/css21_dev/html4/background-color-088.htm and tests/wpt/css-tests/css21_dev/html4/background-color-089.htm (integers and percentages, respectively).

dzbarsky:
Fixes #76

[Review on Reviewable](https://reviewable.io/reviews/servo/rust-cssparser/83)


dzbarsky:
Here's what Gecko does: http://mxr.mozilla.org/mozilla-central/source/layout/style/nsCSSParser.cpp#6441

SimonSapin:
I’ve written in https://github.com/servo/rust-cssparser/issues/76#issuecomment-115269254 why R G and B channels are currently not clamped. My understanding of the spec is that they shouldn’t. I see that Gecko does it, but why?

PR #85, opened 2015-08-09
Add HeapSizeOf support.

notriddle:
Part of servo/heapsize#5.

[Review on Reviewable](https://reviewable.io/reviews/servo/rust-cssparser/85)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -9,6 +9,9 @@
 #![cfg_attr(feature = "serde-serialization", feature(custom_derive))]
 #![cfg_attr(feature = "serde-serialization", feature(plugin))]
 #![cfg_attr(feature = "serde-serialization", plugin(serde_macros))]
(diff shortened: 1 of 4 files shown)


SimonSapin:
So that this code is built on CI, please add a line in the `script` section of `.travis.yml` that is something like this (untested):
```yaml
   - ([ $TRAVIS_RUST_VERSION != nightly ] || cargo test --features heap_size)

File: src/lib.rs [hunk 1]
```diff
@@ -83,6 +84,8 @@ pub use nth::parse_nth;
 pub use serializer::{ToCss, CssStringWriter, serialize_identifier, serialize_string};
 pub use parser::{Parser, Delimiter, Delimiters, SourcePosition};
 
+#[cfg(feature = "heap_size")] known_heap_size!(0, Color, RGBA);
```
jdm: Why use this over deriving the trait on the types?

PR #89, opened 2015-09-09
Add "token serialization types".

SimonSapin:
Servo will use this for custom property values which are conceptually sequences of tokens, but are represented in memory as strings. When concatenating such strings, an empty comment `/**/` sometimes needs to be inserted so that two tokens are not reparsed as one.

r? @pcwalton

[Review on Reviewable](https://reviewable.io/reviews/servo/rust-cssparser/89)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 
 name = "cssparser"
-version = "0.3.8"
+version = "0.3.9"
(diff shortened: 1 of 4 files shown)


mbrubeck:
Reviewed 4 of 4 files at r1.
Review status: :shipit: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/servo/rust-cssparser/89)

PR #90, opened 2015-10-09
Treat url(<string>) as a normal functions, per spec change.

SimonSapin:
Only unquoted URLs are special tokens now. Use `Parser::expect_url`.

This is a [breaking-change]. The version number was incremented accordingly.

This change will help with https://github.com/servo/servo/issues/7767

This triggers https://github.com/rust-lang/rust/issues/28934 and fails to build in the current Rust nightly, but works fine in the Rust version that Servo currently use. Hopefully that rustc bug will be fixed before we need to upgrade Rust in Servo.

r? @mbrubeck

[Review on Reviewable](https://reviewable.io/reviews/servo/rust-cssparser/90)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 
 name = "cssparser"
-version = "0.3.9"
+version = "0.4.0"
 authors = [ "Simon Sapin <simon.sapin@exyr.org>" ]
 
 description = "Rust implementation of CSS Syntax Level 3"
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -331,7 +331,7 @@ impl<'i, 't> Parser<'i, 't> {
     /// This can help tell e.g. `color: green;` from `color: green 4px;`
     #[inline]
     pub fn parse_entirely<F, T>(&mut self, parse: F) -> Result<T, ()>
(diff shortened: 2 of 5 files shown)


mbrubeck:
Reviewed 5 of 5 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks pending.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/servo/rust-cssparser/90)

mbrubeck:
We'll need to disable the nightly channel in Travis temporarily, or pin it to a specific nightly.

SimonSapin:
Let’s wait and see if I want to land the corresponding Servo change before the rustc bug is fixed.

SimonSapin:
The Rust bug is fixed in nightly, but beta shipped with it: https://github.com/rust-lang/rust/pull/29188#issuecomment-152580336

jdm:
Failed on the beta compiler as expected.

SimonSapin:
@bors-servo retry

https://github.com/rust-lang/rust/pull/29188 has been backported to 1.5.0-beta.3: https://github.com/rust-lang/rust/pull/29592

PR #102, opened 2016-03-09
Fix accidentally quadratic algorithm

SimonSapin:
`Tokenize::source_location` takes an "position" counting UTF-8 bytes since the start of the stylesheet and returns a "location" made of a line number and column number. To avoid counting lines from the start every time, it saves some results after each call.

Previously we would only save the position of the last known line break, so calling source_location() for name positions within a single long line would keep searching from that point for the next line break and therefore take O(n²) time.

Very long lines can easily happen when a CSS minifier is used.

We now save the full (line, column) result from the last call.

See https://github.com/servo/servo/issues/9897#issuecomment-194034456 and https://github.com/servo/servo/issues/9897#issuecomment-194167211

r? @jdm

---
This change is [Review on Reviewable](https://reviewable.io/reviews/servo/rust-cssparser/102)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,25 +16,13 @@ license = "MPL-2.0"
 rustc-serialize = "0.3"
 tempdir = "0.3"
 
-[dependencies.serde]
-version = ">=0.6.6, <0.8"
-optional = true
-
-[dependencies.serde_macros]
-version = ">=0.6.5, <0.8"
-optional = true
-
-[dependencies.heapsize]
-version = ">=0.1.1, <0.4.0"
-optional = true
-
-[dependencies.heapsize_plugin]
-version = "0.1.0"
-optional = true
-
 [dependencies]
 encoding = "0.2"
+heapsize = {version = ">=0.1.1, <0.4.0", optional = true}
+heapsize_plugin = {version = "0.1.0", optional = true}
 matches = "0.1"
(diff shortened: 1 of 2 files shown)


src/tests.rs
```diff
@@ -409,7 +409,7 @@ fn line_numbers() {
     assert_eq!(input.next_including_whitespace(), Ok(Token::Ident(Borrowed("foo"))));
     assert_eq!(input.current_source_location(), SourceLocation { line: 1, column: 4 });
     assert_eq!(input.next_including_whitespace(), Ok(Token::WhiteSpace(" ")));
-    assert_eq!(input.current_source_location(), SourceLocation { line: 1, column: 5 });
+//    assert_eq!(input.current_source_location(), SourceLocation { line: 1, column: 5 });
```
Ms2ger: Remove or keep, but don't comment it out.

SimonSapin: Ah, yes, I didn’t mean to commit that. Reverted.

jdm:
@bors-servo: r+
Thanks!

SimonSapin:
Rebased.

@bors-servo r=jdm

Previous failure was `~/rust-installer/rustup.sh` exiting with 1 for no apparent reason.

SimonSapin:
Published v0.5.5 on crates.io

PR #103, opened 2016-03-16
Test serde-serialization feature on Travis CI.

frewsxcv:
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-cssparser/103)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ script:
   - cargo build --verbose
   - cargo test --verbose
   - cargo doc --verbose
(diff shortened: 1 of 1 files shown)


SimonSapin:
I don’t remember if cssparser has had this before, but it was deliberate to either remove it or not add it.

Serde’s dependencies are often broken by libsyntax changes which would cause Travis to fail in this repository for something that can’t be fixed in this repository. On the other hand, the only code in this repository is `#[derive]` so it’s very hard to have a bug there. This PR would almost never help us catch real bugs in this repository.

frewsxcv:
Fair enough!

PR #104, opened 2016-05-04
New characters escaped in serialize_name

mskrzypkows:
Added escape of NULL (U+0000) char and escape for characters
in the range [\1-\1f] (U+0001 to U+001F) or U+007F
according to https://drafts.csswg.org/cssom/#serialize-an-identifier
Needed for proper CSS tests https://github.com/servo/servo/issues/10685

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-cssparser/104)

### DIFF
--- a/src/serializer.rs
+++ b/src/serializer.rs
@@ -196,14 +196,14 @@ fn serialize_name<W>(value: &str, dest: &mut W) -> fmt::Result where W:fmt::Writ
         let escaped = match b {
             b'0'...b'9' | b'A'...b'Z' | b'a'...b'z' | b'_' | b'-' => continue,
             _ if !b.is_ascii() => continue,
-            b'\n' => Some("\\A "),
-            b'\r' => Some("\\D "),
-            b'\x0C' => Some("\\C "),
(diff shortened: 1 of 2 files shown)


nox:
The code looks correct, but could you add some unit tests please?

mskrzypkows:
OK I'll add, forgot about it.

SimonSapin:
Please add unit tests in `src/tests.rs` of the form:

```rust
#[test]
fn identifier_serialization() {
    assert_eq!(Token::Ident("\0").to_css_string(), "\u{FFFD}");
    assert_eq!(Token::Ident("a\0").to_css_string(), "a\u{FFFD}");
    // ...
}
```

with one `assert_eq!` for each `assert_equals` in https://github.com/servo/servo/blob/master/tests/wpt/css-tests/cssom-1_dev/html/escape.htm, except for "Number of arguments", "String conversion", and "Surrogates" ones. (But add one with `"\u{1D306}"` instead of surrogates.)

mskrzypkows:
I've added tests.

SimonSapin:
Great, thanks!

@bors-servo r+

PR #115, opened 2017-02-02
Preconvert in Color::parse to avoid codesize impact

Manishearth:
Do not land.

We currently ascii-lowercase each time. We shouldn't do that in large matches.

We should probably have a COW-returning method instead of an unconditional allocation here; so that we retain the speed of not allocating in the common case.

cc @mbrubeck does this improve codesize?

r? @SimonSapin

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-cssparser/115)

### DIFF
--- a/src/color.rs
+++ b/src/color.rs
@@ -135,7 +135,8 @@ fn rgba(red: f32, green: f32, blue: f32, alpha: f32) -> Result<Color, ()> {
 /// (For example, the value of an `Ident` token is fine.)
 #[inline]
 pub fn parse_color_keyword(ident: &str) -> Result<Color, ()> {
-    match_ignore_ascii_case! { ident,
+    use std::ascii::AsciiExt;
+    match &*ident.to_ascii_lowercase() {
         "black" => rgb(0., 0., 0.),
         "silver" => rgb(192., 192., 192.),
         "gray" => rgb(128., 128., 128.),


SimonSapin:
I don’t think repeatedly lowercasing is what hurts code size. (Though it might hurt runtime perf.) It’s rather generating code for each color keyword. I think having a static table and a loop that uses `eq_ignore_ascii_case` can also reduce code size. I’ll make a PR now.

Manishearth:
Yeah, that's what dmajor suggested. I also wanted to try this because this may improve perf (once used with COW, that is)

SimonSapin:
Here is the static table: https://github.com/servo/rust-cssparser/compare/color-keyword-static-table

I’ve written the `Cow` lower case thing before but didn’t end up using it. Here it is:

```rust
/// Like AsciiExt::to_ascii_lowercase, but avoids allocating when the input is already lower-case.
pub fn cow_into_ascii_lowercase<'a, S: Into<Cow<'a, str>>>(s: S) -> Cow<'a, str> {
    let mut cow = s.into();
    match cow.bytes().position(|byte| byte >= b'A' && byte <= b'Z') {
        Some(first_uppercase) => cow.to_mut()[first_uppercase..].make_ascii_lowercase(),
        None => {}
    }
    cow
}
```

Manishearth:
Should we merge that branch instead?

Manishearth:
r=me if you want to land it

PR #119, opened 2017-02-14
Add comments to justify `unsafe` blocks.

SimonSapin:
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-cssparser/119)

### DIFF
--- a/src/tokenizer.rs
+++ b/src/tokenizer.rs
@@ -576,6 +576,7 @@ fn consume_string<'a>(tokenizer: &mut Tokenizer<'a>, single_quote: bool) -> Toke
 fn consume_quoted_string<'a>(tokenizer: &mut Tokenizer<'a>, single_quote: bool)
                              -> Result<Cow<'a, str>, ()> {
     tokenizer.advance(1);  // Skip the initial quote
+    // start_pos is at code point boundary, after " or '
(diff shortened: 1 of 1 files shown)


src/tokenizer.rs
```diff
@@ -1039,11 +1070,8 @@ fn consume_hex_digits<'a>(tokenizer: &mut Tokenizer<'a>) -> (u32, u32) {
 
 // Same constraints as consume_escape except it writes into `bytes` the result
 // instead of returning it.
-//
-// TODO: This could be made more efficient with char::encode_utf8, I guess.
 fn consume_escape_and_write(tokenizer: &mut Tokenizer, bytes: &mut Vec<u8>) {
-    use std::io::Write;
-    write!(bytes, "{}", consume_escape(tokenizer)).unwrap();
+    bytes.extend(consume_escape(tokenizer).encode_utf8(&mut [0; 4]).as_bytes())
```
SimonSapin: Note that `encode_utf8` is stable since Rust 1.15.

I’m not convinced this requires a breaking version number increment, but regardless so we’re doing one in https://github.com/servo/rust-cssparser/pull/118.

emilio:
@bors-servo r+

Yeah, let's roll that change into the colors thing.

PR #121, opened 2017-02-17
color: Support CSS Color Level 4 rgb & hsl syntax (#113)

### LINKED ISSUE #113: Support rgb and hsl new syntax in CSS Color Level 4
That is, to
* make `rgb` an alias of `rgba`
* make `hsl` an alias of `hsla`
* make alpha channel optional for all functions above
* accept whitespace splited form

https://drafts.csswg.org/css-color-4/#rgb-functions
https://drafts.csswg.org/css-color-4/#the-hsl-notation

manuel-woelker:
Fixes #113 - Feedback welcome!

If I interpret the spec correctly, in the whitespace-split version the alpha channel is separated with a `/`.
I also added support for angle values (deg/grad/rad/turn) in the hue position.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-cssparser/121)


src/color.rs
```diff
@@ -386,47 +387,84 @@ fn clamp_f32(val: f32) -> u8 {
 
 #[inline]
 fn parse_color_function(name: &str, arguments: &mut Parser) -> Result<Color, ()> {
-    let (is_rgb, has_alpha) = match_ignore_ascii_case! { name,
-        "rgba" => (true, true),
-        "rgb" => (true, false),
-        "hsl" => (false, false),
-        "hsla" => (false, true),
+    let is_rgb = match_ignore_ascii_case! { name,
+        "rgba" => true,
```
emilio: I believe `match_ignore_ascii_case` now supports more general patterns, so you can write it down as:

```rust
let is_rgb = match_ignore_ascii_case! { name,
    "rgba" | "rgb" => true,
    "hsl" | "hsla" => false,
    _ => return Err(()),
};
```

src/color.rs
```diff
@@ -386,47 +387,84 @@ fn clamp_f32(val: f32) -> u8 {
...
-        let saturation = try!(arguments.expect_percentage()).max(0.).min(1.);
-        try!(arguments.expect_comma());
+        let saturation = match try!(arguments.next()) {
+            Token::Percentage(ref v) => v.unit_value,
+            Token::Comma => {
+                uses_commas = true;
+                try!(arguments.expect_percentage())
+            }
+            _ => return Err(())
+        }.max(0.).min(1.);
```
emilio: I think this bit is easy to miss. What about splitting it into other line?

```rust
let saturation = match ... {
};
let saturation = saturation.max(0.).min(1.);
```

src/color.rs
```diff
@@ -386,47 +387,84 @@ fn clamp_f32(val: f32) -> u8 {
...
                 blue = clamp_f32(try!(arguments.expect_percentage()));
             }
             _ => return Err(())
         };
     } else {
-        let hue_degrees = try!(arguments.expect_number());
+        let hue_degrees = match try!(arguments.next()) {
+                Token::Number(NumericValue { value: v, .. }) => v,
+                Token::Dimension(NumericValue { value: v, .. }, unit) => {
+                    match &*unit {
```
emilio: This needs to use case insensitive matching, please use `match_ignore_ascii_case!`. Spec: https://drafts.csswg.org/css-values/#dimensions

> Like keywords, unit identifiers are ASCII case-insensitive

emilio: Also, can you drop a link here to: https://drafts.csswg.org/css-values/#angles?

src/color.rs
```diff
@@ -386,47 +387,84 @@ fn clamp_f32(val: f32) -> u8 {
...
+                    try!(arguments.expect_comma());
+                }
                 blue = clamp_f32(try!(arguments.expect_percentage()));
             }
             _ => return Err(())
         };
     } else {
-        let hue_degrees = try!(arguments.expect_number());
+        let hue_degrees = match try!(arguments.next()) {
+                Token::Number(NumericValue { value: v, .. }) => v,
```
emilio: nit: Indentation here is off, please indent four spaces from the previous letter, here and below.

src/color.rs
```diff
@@ -449,12 +487,38 @@ fn parse_color_function(name: &str, arguments: &mut Parser) -> Result<Color, ()>
...
+    };
+
+    /*
     let alpha = if has_alpha {
         try!(arguments.expect_comma());
         clamp_f32(try!(arguments.expect_number()))
     } else {
         255
-    };
+    };*/
```
emilio: Please remove?

src/color.rs
```diff
@@ -386,47 +387,84 @@ fn clamp_f32(val: f32) -> u8 {
...
+        "hsl" => false,
+        "hsla" => false,
         _ => return Err(())
     };
 
     let red: u8;
     let green: u8;
     let blue: u8;
+    let mut uses_commas = false;
     if is_rgb {
```
emilio: I believe that, now that this function is basically a big:

```
if is_rgb {
    // ...
} else {
    // ...
}

try_parse_alpha()
```

we can split it the giant `if` into two functions to get the relevant rgb components:

```rust
fn parse_rgb_components_rgb(arguments: &mut Parser) -> Result<(u8, u8, u8, bool), ()>;
fn parse_rgb_components_hsl(arguments: &mut Parser) -> Result<(u8, u8, u8, bool), ()>;
```

This would look something like:

```rust
let (red, green, blue, uses_commas) = if is_rgb {
    parse_rgb_components_rgb(arguments)?
} else {
    parse_rgb_components_hsl(arguments)?
};

// Code for parsing the optional alpha value.
```

What do you think? I believe this should make this function much easier to read and understand.

src/color.rs
emilio: Also, please add a test for this as you see fit :)

emilio:
Hi, are you still able to work on this? Happy to take over the remaining work here.

Thanks!

manuel-woelker:
Sorry, I missed the last update. Thanks for the review. I'll try to implement the changes this week.

emilio:
No worries, and thanks!

manuel-woelker:
I rebased onto master and updated the PR. Let me know if there is anything else or you want me to squash the commits.

src/color.rs
```diff
@@ -524,13 +436,114 @@ fn parse_color_function(name: &str, arguments: &mut Parser) -> Result<Color, ()>
...
+    let mut uses_commas = false;
+    // Hue given as an angle
+    // https://drafts.csswg.org/css-values/#angles
+    let hue_degrees = match try!(arguments.next()) {
+        Token::Number(NumericValue { value: v, .. }) => v,
+        Token::Dimension(NumericValue { value: v, .. }, unit) => {
+            match_ignore_ascii_case! { &*unit,
+                "deg" => v,
+                "grad" => v * 360. / 400.,
+                "rad" => v * 360. / (2. * PI as f32),
```
emilio: One last nit: Let's use `std::f32::consts::PI` instead of the `f64` version unless there's a good reason for that.

manuel-woelker:
Done, thanks for the feedback!

SimonSapin:
Thank you for working on this!

@bors-servo r=emilio,SimonSapin

PR #122, opened 2017-02-25
Make match_ignore_ascii_case more efficient, add ascii_case_insensitive_phf_map

SimonSapin:
This improves the performance of `match_ignore_ascii_case!` by replacing calls to `str::eq_ignore_ascii_case` with plain string equality. The input string is converted to ASCII lower-case once, using a stack allocated buffer. A `proc_macro_derive` is used internally to ensure that string patterns are already lower-case, and to computed the size of the buffer (the length of the longest pattern).

This should improve runtime performance, it the amount of generated MIR or LLVM IR in release mode still looks proportional to the number of patterns, so this by itself won’t help much with the code bloat in `parse_color_keyword`.

To deal with that, this PR also adds a `ascii_case_insensitive_phf_map!` macro that reuses the same stack-allocated buffer mechanism to lower-case an input string, and combines it with a [`phf`](https://github.com/sfackler/rust-phf) static hash map.

The two macros are similar but make different treadoffs. PHF probably generates faster and less bloated code, but the map’s values need to be given as a string that contains Rust syntax, due to limitations of procedural macros in current stable Rust. On the other hand, generating a `match` expression allows using control flow statements like `return` or `continue` in its match arms.

Fixes #115.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-cssparser/122)


SimonSapin:
r? @Manishearth, CC @nox

Do not merge just yet, I’ll make sure Servo can build with this first :)

SimonSapin:
@bors-servo retry

Job timed out.

SimonSapin:
Not sure what homu is doing, https://travis-ci.org/servo/rust-cssparser/builds/205343338 is green.

PR #123, opened 2017-02-28
Rewrite procedural macro based on the enum discriminant trick

SimonSapin:
#122 introduced a procedural component to some `macro_rules` macros by abusing `derive` on a dummy type declaration. (Custom `derive` is available on stable Rust 1.15, but functional procedural macros are not.) It used an attribute to pass data to the procedural macro, but this data is limited to string literals in Rust 1.15.

We can improve on this with an idea from @dtolnay:

```rust
#[derive(Something)]
enum Dummy {
    Input = (0, stringify!( ... )).0
}
```

Instead of a custom attribute, we use an enum with an explicit discriminant (because that’s the only case where a type definition can contain a (const) expression), together with accessing the first field of a literal tuple to ignore other fields.

Instead of just string literals, we can accept arbitrary tokens that `stringify!` will later be expanded into something parseable in a const expression context. (`stringify!` doesn’t help in attributes because it is not yet expanded by the time anything other than string literals is rejected.)

The details of this hack are hidden away into a new crate, called [`procedural-masquerade`](https://docs.rs/procedural-masquerade/).

----

This enables concrete benefits for users of cssparser:

* `match_ignore_ascii_case!` now supports the full syntax of `match` expressions, including alternatives `pattern | other_pattern`, bindings `name @ pattern`, and guards `pattern if condition`.
* Map values in `asci_case_insensitive_phf_map!` are now given literally instead of as string literals containing Rust syntax.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-cssparser/123)


SimonSapin:
r? @nox

Do not merge yet, I’ll port Servo and figure out how to land it without breaking Firefox too much…

nox:
r=me

PR #124, opened 2017-03-03
Don't heap-allocate a tokenizer for top-level parsing.

emilio:
r? @SimonSapin

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-cssparser/124)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 
 name = "cssparser"
-version = "0.12.0"
+version = "0.12.1"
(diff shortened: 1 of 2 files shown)


SimonSapin:
IIRC when I wrote this I did not expect creating new parsers to be hot. Is it for CSSOM setters?

SimonSapin:
Ah you already replied yes on IRC :)

@bors-servo r+

emilio:
@bors-servo r-

I'll update the MaybeOwned docs.

SimonSapin:
(MaybeOwned is not public, but yeah)

@bors-servo r+

PR #127, opened 2017-03-09
macros: Make bindings work as expected.

### LINKED ISSUE #126: `match_ignore_ascii_case` fails to return value after some characters
Simple test case:
```patch
diff --git a/src/tests.rs b/src/tests.rs
index 8419943..fe19381 100644
--- a/src/tests.rs
+++ b/src/tests.rs
@@ -843,3 +843,21 @@ fn procedural_masquerade_whitespace() {
         _ => panic!("4"),
     }
 }
+
+#[test]
+fn test_length() {
+    match_ignore_ascii_case! { "fo",
+        "bar" => {},
+        value => assert_eq!(value, "fo"),
+    }
+
+    match_ignore_ascii_case! { "foo",
+        "bar" => {},
+        value => assert_eq!(value, "foo"),
+    }
+
+    match_ignore_ascii_case! { "fooo",
+        "bar" => {},
+        value => assert_eq!(value, "fooo"),
+    }
+}
```
In this function, first two `match_ignore_ascii_case` passes but in the third macro, value returns "A". I have the same problem with this macro in https://github.com/servo/servo/pull/15813 But it breaks after 11 characters.

cc @SimonSapin

emilio:
Fixes #126 

This change is [Reviewable](https://reviewable.io/reviews/servo/rust-cssparser/127)


emilio:
Note that this is somewhat a wallpaper-ish patch, since the string would be lowercased if its length is less that the bindings.

We can disable bindings altogether otherwise I guess.

SimonSapin:
I think you’re right: we can’t have all three of lowercasing into a buffer on the stack to avoid allocation, support bindings in `match` arms, and get exactly the behavior of `match &*input.to_ascii_lowercase() { /* … */ }` (other than allocation). I’m also tempted to drop support for bindings in match arms.

@canaltinova What was your use case for bindings? Can you write your code without? Maybe replacing them like this:

```diff
-match_ignore_ascii_case! { &expect_keyword(),
+let keyword = expect_keyword();
+match_ignore_ascii_case! { &keyword,
     "foo" => Foo,
     "bar" => Bar,
-    keyword => Other(keyword.to_owned()),
+    _ => Other(keyword.to_owned()),
 }
```

canaltinova:
will-change property accepts `<custom-ident>`. I was using this to convert to lowercase and exclude some unacceptable identifiers. Here's how I use this macro:
```rust
match_ignore_ascii_case! { &i.expect_ident()?,
    "will-change" | "none" | "all" | "auto" |
    "initial" | "inherit" | "unset" => Err(()),
    ident  => {
        Ok((Atom::from(ident)))
    }
}
```
Ah, `<custom-ident>` is actually case sensitive. I think I can live with that 😃

SimonSapin:
Ah, yes. If you want to get the lower-case version of an arbitrary string (of unbounded length), not just match it against a set of pre-defined patterns, then you should use `s.to_ascii_lowercase()` + `match` instead of `match_ignore_ascii_case!`.

SimonSapin:
Thanks for the PR @emilio but I think I’d rather remove support for match bindings. With this PR the bound value would sometimes be lower-cased and sometimes not, depending on its length. Closing, I’ll work on another PR now.

PR #129, opened 2017-03-19
Character between U+0001 ~ U+001F in string should be escaped when serialized

ak1t0:
Fixes [servo/servo#15947](https://github.com/servo/servo/issues/15947)

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-cssparser/129)

### DIFF
--- a/src/serializer.rs
+++ b/src/serializer.rs
@@ -250,12 +250,17 @@ impl<'a, W> fmt::Write for CssStringWriter<'a, W> where W: fmt::Write {
     fn write_str(&mut self, s: &str) -> fmt::Result {
         let mut chunk_start = 0;
         for (i, b) in s.bytes().enumerate() {
+            let string;
             let escaped = match b {
                 b'"' => "\\\"",
                 b'\\' => "\\\\",
                 b'\n' => "\\A ",
                 b'\r' => "\\D ",
-                b'\x0C' => "\\C ",
+                b'\0' => "\u{FFFD}",
+                x if (x >= b'\x01' && x <= b'\x1F') || x == b'\x7F' => {
+                    string = format!("\\{:x} ", x);
+                    &string
+                },
                 _ => continue,
             };
             try!(self.inner.write_str(&s[chunk_start..i]));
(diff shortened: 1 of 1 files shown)


SimonSapin:
Reviewed 1 of 1 files at r1.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[src/serializer.rs, line 260 at r1](https://reviewable.io:443/reviews/servo/rust-cssparser/129#-KfbAu_njWQWxqP6uEil:-KfbAu_njWQWxqP6uEim:bpcktui) ([raw file](https://github.com/servo/rust-cssparser/blob/185b35de6ab8c8cdfaa075d21f504fb5df2348f6/src/serializer.rs#L260)):*
> ```Rust
>                 b'\r' => "\\D ",
>                 b'\0' => "\u{FFFD}",
>                 x if (x >= b'\x01' && x <= b'\x1F') || x == b'\x7F' => {
> ```

A range pattern here would look nicer than a guard:`b'\x01'...b'\x1F' | b'\x7F'`

---

*[src/serializer.rs, line 261 at r1](https://reviewable.io:443/reviews/servo/rust-cssparser/129#-KfbBBMHpvRYYMTLA_Ct:-KfbBBMHpvRYYMTLA_Cu:b-hx5eso) ([raw file](https://github.com/servo/rust-cssparser/blob/185b35de6ab8c8cdfaa075d21f504fb5df2348f6/src/serializer.rs#L261)):*
> ```Rust
>                 b'\0' => "\u{FFFD}",
>                 x if (x >= b'\x01' && x <= b'\x1F') || x == b'\x7F' => {
>                     string = format!("\\{:x} ", x);
> ```

`format!` does memory allocation in order to return a string, which can be costly. To avoid it, change the type of this `match` expression from `&str` to `Option<&str>` (so wrapping str literals in `Some( )` in other branches) where `None` indicates that the byte needs to be hexadecimal-escaped. Below, use `try!(write!(self.inner, "\\{:x}", b));` in that case to write directly to the output stream without an intermediate `String`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/rust-cssparser/129)*

ak1t0:
I fixed that as it was pointed out to me.

SimonSapin:
Thank you! Looks good.

@bors-servo r+

---

Reviewed 1 of 1 files at r2, 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/rust-cssparser/129#-:-KfjmvgXggxeV6cAKODW:b2hk5cx)*

PR #130, opened 2017-04-03
Stop using temp files and an external program for diffs in test failures

SimonSapin:
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-cssparser/130)


emilio:
Quite nicer!

@bors-servo r+

PR #132, opened 2017-04-07
Numeric faster

emilio:
The following commit:

 * Removes char logic from consume_numeric.
 * Rearranges exponent parsing to look at one char in the common case
   (presumably not a big deal).

This makes the benchmark added in the first commit go from:

  test tests::numeric      ... bench:  48,221,308 ns/iter (+/- 3,356,881)

to:

  test tests::numeric      ... bench:  43,656,802 ns/iter (+/- 1,323,570)

on my machine.

Which is pretty much a micro-optimization, but perhaps it's worth given how
common these values are.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-cssparser/132)

### DIFF
--- a/src/tokenizer.rs
+++ b/src/tokenizer.rs
@@ -750,6 +750,24 @@ fn consume_name<'a>(tokenizer: &mut Tokenizer<'a>) -> Cow<'a, str> {
     )
 }
 
+fn byte_to_hex_digit(b: u8) -> Option<u32> {
+    Some(match_byte! { b,
+        b'0' ... b'9' => { b - b'0' },
+        b'a' ... b'f' => { b - b'a' + 10 },
+        b'A' ... b'F' => { b - b'A' + 10 },
+        _ => {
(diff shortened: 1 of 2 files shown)


src/tokenizer.rs
```diff
@@ -750,6 +750,24 @@ fn consume_name<'a>(tokenizer: &mut Tokenizer<'a>) -> Cow<'a, str> {
     )
 }
 
+fn byte_to_hex_digit(b: u8) -> Option<u32> {
+    Some(match_byte! { b,
+        b'0' ... b'9' => { b - b'0' },
+        b'a' ... b'z' => { b - b'a' + 10 },
+        b'A' ... b'Z' => { b - b'A' },
```
jdm: This is an unusual definition of hexadecimal.

emilio: Whoopsies, indeed :)

PR #139, opened 2017-04-16
Let define_proc_macros! handle early returns.

purpleposeidon:
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-cssparser/139)


SimonSapin:
Ah, I didn’t think of that. Looks good, thanks!

PR #142, opened 2017-05-04
Use Gecko's behavior for calculating percent colors

Manishearth:
https://bugzilla.mozilla.org/show_bug.cgi?id=1340484

This change is [Reviewable](https://reviewable.io/reviews/servo/rust-cssparser/142)


emilio:
Needs tests update.

Manishearth:
Gecko also computes `rgb(254.5, 254.6, 254.4)` as `rgb(254, 254, 254)`, i.e. it rounds float colors down. It's really nonsensical to specify float colors anyway but I made us match Gecko here too.

Manishearth:
(tests updated)

SimonSapin:
> It's really nonsensical to specify float colors anyway

Only in impls that store components as `u8` :) I’m told that Safari uses `f64`. (They support high-gamut displays where 8 bits is insufficient. I don’t know if 64 useful or excessive.)

---

Reviewed 5 of 5 files at r1.

---

*[src/color.rs, line 406 at r1](https://reviewable.io:443/reviews/servo/rust-cssparser/142#-KjsV7jT3U2bEOYvFr4s:-KjsV7jUZ5_0kdhfT4MG:b-vb0je2) ([raw file](https://github.com/servo/rust-cssparser/blob/49a2e6330f55d192e6ef36b575bc6a9a041f6b7f/src/color.rs#L406)):*
> ```Rust
> }
> 
> fn clamp_256_f32(val: f32) -> u8 {
> ```

Maybe this function should be renamed now?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/rust-cssparser/142#-:-KjsVQdtJNTNgDcxILy1:b-4kte2l)*

SimonSapin:
r=me with that

PR #143, opened 2017-05-05
Return more specific errors from parsing APIs

jdm:
This shouldn't be merged until the related Servo and Gecko work is complete, but I figured I would put this up for feedback.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-cssparser/143)

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -21,6 +21,16 @@ pub struct SourcePosition {
     at_start_of: Option<BlockType>,
 }
 
+///
+#[derive(Clone, Debug, PartialEq)]
+pub enum ParseError<'a> {
+    ///
+    UnexpectedToken(Token<'a>),
+    ///
+    ExpectedToken(Token<'a>),
+    ///
+    UnexpectedEof,
+}
 
 /// Like std::borrow::Cow, except the borrowed variant contains a mutable
 /// reference.
@@ -188,13 +198,11 @@ impl<'i, 't> Parser<'i, 't> {
     ///
     /// This ignores whitespace and comments.
     #[inline]
-    pub fn expect_exhausted(&mut self) -> Result<(), ()> {
+    pub fn expect_exhausted(&mut self) -> Result<(), ParseError<'i>> {
         let start_position = self.position();
         let result = match self.next() {
-            Err(()) => Ok(()),
-            Ok(_) => {
-                Err(())
-            }
+            Err(_) => Ok(()), //XXXjdm swallowing non-UnexpectedEof errors seems wrong, but tests fail
+            Ok(t) => Err(ParseError::UnexpectedToken(t)),
         };
         self.reset(start_position);
         result
@@ -298,7 +306,7 @@ impl<'i, 't> Parser<'i, 't> {
     /// See the `Parser::parse_nested_block` method to parse the content of functions or blocks.
     ///
     /// This only returns a closing token when it is unmatched (and therefore an error).
-    pub fn next(&mut self) -> Result<Token<'i>, ()> {
+    pub fn next(&mut self) -> Result<Token<'i>, ParseError<'i>> {
         loop {
             match self.next_including_whitespace_and_comments() {
                 Ok(Token::WhiteSpace(_)) | Ok(Token::Comment(_)) => {},
@@ -308,7 +316,7 @@ impl<'i, 't> Parser<'i, 't> {
     }
 
     /// Same as `Parser::next`, but does not skip whitespace tokens.
-    pub fn next_including_whitespace(&mut self) -> Result<Token<'i>, ()> {
+    pub fn next_including_whitespace(&mut self) -> Result<Token<'i>, ParseError<'i>> {
         loop {
             match self.next_including_whitespace_and_comments() {
                 Ok(Token::Comment(_)) => {},
@@ -323,14 +331,15 @@ impl<'i, 't> Parser<'i, 't> {
     /// where comments are preserved.
     /// When parsing higher-level values, per the CSS Syntax specification,
     /// comments should always be ignored between tokens.
-    pub fn next_including_whitespace_and_comments(&mut self) -> Result<Token<'i>, ()> {
+    pub fn next_including_whitespace_and_comments(&mut self) -> Result<Token<'i>, ParseError<'i>> {
         if let Some(block_type) = self.at_start_of.take() {
             consume_until_end_of_block(block_type, &mut *self.tokenizer);
         }
-        if self.stop_before.contains(Delimiters::from_byte(self.tokenizer.next_byte())) {
-            return Err(())
+        let byte = self.tokenizer.next_byte();
+        if self.stop_before.contains(Delimiters::from_byte(byte)) {
+            return Err(ParseError::UnexpectedToken(Token::Delim(byte.unwrap_or(0) as char)))
         }
-        let token = try!(self.tokenizer.next());
+        let token = try!(self.tokenizer.next().map_err(|()| ParseError::UnexpectedEof));
         if let Some(block_type) = BlockType::opening(&token) {
             self.at_start_of = Some(block_type);
         }
@@ -342,8 +351,8 @@ impl<'i, 't> Parser<'i, 't> {
     ///
     /// This can help tell e.g. `color: green;` from `color: green 4px;`
     #[inline]
-    pub fn parse_entirely<F, T>(&mut self, parse: F) -> Result<T, ()>
-    where F: FnOnce(&mut Parser<'i, 't>) -> Result<T, ()> {
+    pub fn parse_entirely<F, T>(&mut self, parse: F) -> Result<T, ParseError<'i>>
+    where F: FnOnce(&mut Parser<'i, 't>) -> Result<T, ParseError<'i>> {
         let result = parse(self);
         try!(self.expect_exhausted());
         result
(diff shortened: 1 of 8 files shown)


jdm:
Since https://github.com/servo/servo/pull/16752 is ready for review, please review this!

src/parser.rs
```diff
@@ -188,13 +227,12 @@ impl<'i, 't> Parser<'i, 't> {
...
     #[inline]
-    pub fn expect_exhausted(&mut self) -> Result<(), ()> {
+    pub fn expect_exhausted(&mut self) -> Result<(), BasicParseError<'i>> {
         let start_position = self.position();
         let result = match self.next() {
-            Err(()) => Ok(()),
-            Ok(_) => {
-                Err(())
-            }
+            //FIXME: swallowing non-UnexpectedEof errors seems wrong, but tests fail otherwise.
```
emilio: Huh... Could you file an issue with more details, and perhaps an example, and reference it here, just to see if it's something expected or not? Which other errors are reported?

(also, uber-nit: a space after `//`)

src/rules_and_declarations.rs
```diff
@@ -52,6 +52,9 @@ pub trait DeclarationParser {
     /// The finished representation of a declaration.
     type Declaration;
 
+    ///
```
emilio: Missing docs? :)

src/rules_and_declarations.rs
```diff
@@ -89,6 +93,9 @@ pub trait AtRuleParser {
     /// The finished representation of an at-rule.
     type AtRule;
 
+    ///
+    type Error;
```
emilio: Please document it too

src/rules_and_declarations.rs
```diff
@@ -266,8 +284,8 @@ pub struct RuleListParser<'i: 't, 't: 'a, 'a, P> {
 }
 
 
-impl<'i: 't, 't: 'a, 'a, R, P> RuleListParser<'i, 't, 'a, P>
-where P: QualifiedRuleParser<QualifiedRule = R> + AtRuleParser<AtRule = R> {
+impl<'i: 't, 't: 'a, 'a, R, P, E> RuleListParser<'i, 't, 'a, P>
```
emilio: Wow, those are a lot of lifetimes.

src/parser.rs
jdm: Filed #144.

jdm:
Review comments addressed. @SimonSapin when do you think you will be able to look at this?

src/parser.rs
emilio: I took a look at #144, It's legit and we should remove the fixme and perhaps add a comment there.

src/rules_and_declarations.rs
```diff
@@ -157,6 +164,9 @@ pub trait QualifiedRuleParser {
     /// The finished representation of a qualified rule.
     type QualifiedRule;
 
+    ///
```
emilio: This one seems to still be missing. A copy of the other `Error` docs is probably ok.

SimonSapin:
IRC conversation:

> * SimonSapin>	jdm: looking at https://github.com/servo/rust-cssparser/pull/143 , I don’t have a precise counter-proposal but I think the set of error types don’t quite make sense
> * jdm>	go on
> * SimonSapin>	jdm: for PreciseParseError should be the basic thing
> * SimonSapin>	jdm: and the generic ParseError::Custom(T) feels like the wrong pattern
> * jdm isn't sure how else to incorporate errors from outside of the library
> * SimonSapin>	jdm: the Basic / Custom enum should not exist, instead things should use From conversions as needed
> * jdm>	conversions into what?
> * jdm>	SimonSapin: keep in mind that BasicParseError exists as a separate type to reduce the number of explicit type annotations that are required
> * jdm>	SimonSapin: when I threaded ParseError<E> through the whole API it because extremely frustrating to use
> * jdm>	so for the APIs that can't return custom types, it's much easier to return a small set of known errors
> * SimonSapin>	jdm: for example instead of returning Result <T, ParseError<'i, E>>, parse_entirely could return Result<T, E> where E: From<RemainingInput>
> * SimonSapin>	or E: From<ParseError>
> * jdm>	see, that would lead to a lot of ambugious types that would need resolving
> * jdm>	and that becomes very gross
> * jdm>	actually I might be misinterpreting that
> * jdm>	SimonSapin: what is RemainingInput?
> * SimonSapin>	jdm: an hypothetical unit type
> * jdm>	hmm
> * jdm>	SimonSapin: ok, I have a minimal form of selectors and style that I can prototype changes to rust-cssparser against
> * jdm>	SimonSapin: so I'll try that out
> * SimonSapin>	jdm: minimal form?
> * jdm>	SimonSapin: yeah; it has parsing code for background-color, custom properties, and something that uses the selectors/ parsing traits
> * jdm>	SimonSapin: rather than having to redo the return types for about 3000 parsing functions
> * SimonSapin>	interesting
> * SimonSapin>	jdm: alternative idea: no generic E, one ParseError type to rule them all. Add variants/fields to it as needed. For example `&'static str` that describes what was expected instead of that unexpected token 
> * jdm>	SimonSapin: could you record this conversation in the PR, please?
> * SimonSapin>	sure
> * jdm>	SimonSapin: I don't think the one type to rule them all makes sense given how many errors will be added to style/
> * SimonSapin>	jdm: I haven’t looked at the servo PR yet. Are there many error types? Why?
> * jdm>	SimonSapin: because we want to provide enough information for gecko to use https://dxr.mozilla.org/mozilla-central/source/dom/locales/en-US/chrome/layout/css.properties
> * jdm>	SimonSapin: the servo PR provides useful context
> * jdm>	(for understanding the rust-cssparser changes)
> * SimonSapin>	jdm: do we need this exact set of error cases, or just providing roughly as useful info?
> * jdm>	SimonSapin: hard to say; I'm aiming for more information than my first attempt could provide.
> * jdm>	which was basically nothing more than "this rule was invalid"

jdm:
I've spent some time trying to integrate your ideas about returning E where `E: From<...>` (I've started out with `From<BasicParseError<'i>>` as something concrete), and I'm really having trouble making any traction. Between issues with incorporating lifetimes (such as the error types in the traits in rules_and_declarations.rs) and type-inference errors when there are too many layers of generics, I don't see a clear path forward. I would like to move forward with the existing design that actually compiles unless you have a concrete counter-proposal.

SimonSapin:
I’m still convince that the generic `enum ParseError<T> { …, Custom(T) }` is not a good pattern, but at the same time if I spent a bunch of time to come up with a concrete counter-proposal I’m likely to hit the same issues as you did and I don’t know that I’ll have any better solution. So let’s call this r+ on the general approach.

---

Reviewed 5 of 7 files at r2, 2 of 2 files at r4.
Review status: all files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[src/parser.rs, line 373 at r4](https://reviewable.io:443/reviews/servo/rust-cssparser/143#-KkR0_K6kX1SCl087LhU:-KkR0_K7SY54USCMfs60:bkxkwya) ([raw file](https://github.com/servo/rust-cssparser/blob/1b8dcea7b0812bfa5dc89b7603515e8cc64e30ba/src/parser.rs#L373)):*
> ```Rust
>         let byte = self.tokenizer.next_byte();
>         if self.stop_before.contains(Delimiters::from_byte(byte)) {
>             return Err(BasicParseError::UnexpectedToken(Token::Delim(byte.unwrap_or(0) as char)))
> ```

Returning `UnexpectedToken` here is incorrect. The point of `self.stop_before` is to provide the abstraction that we’re parsing a subslice of the original stream. Finding the byte/token that marks the end of that sub-stream is the expected end of the stream.

Maybe `UnexpectedEof` should be renamed `EndOfInput`?

---

*[src/parser.rs, line 610 at r4](https://reviewable.io:443/reviews/servo/rust-cssparser/143#-KkR1JxAsopOBpThBLw0:-KkR1JxAsopOBpThBLw1:bf0byod) ([raw file](https://github.com/servo/rust-cssparser/blob/1b8dcea7b0812bfa5dc89b7603515e8cc64e30ba/src/parser.rs#L610)):*
> ```Rust
>         match token {
>             Token::Number(NumericValue { ref int_value, .. }) if int_value.is_some() => {
>                 Ok(int_value.unwrap())
> ```

Please use `int_value: Some(int_value)` in the pattern here to make `unwrap` unnecessary.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/rust-cssparser/143#-:-KkR1zJKp359tJPxeZip:bdh5fx0)*

nox:
I don't like this design either.

What I would have expected is for `cssparser` to only define a basic set of errors, and for the `Parse` trait in Servo to have an associated type `Error`, like Serde's `Serializer` trait.

This way each specific parser in Servo would care only about its specific errors, and we wouldn't need to have all errors possible in Servo in a single huge enum.

SimonSapin:
I suspect that reconciling lifetime parameters with generic error types would require "associated type constructors" https://github.com/rust-lang/rfcs/pull/1598

nox:
I see.

SimonSapin:
@bors-servo r+

---

Reviewed 8 of 8 files at r6.
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/rust-cssparser/143#-:-KkR6FISg4CoUsP-wIS-:bkg9nno)*

PR #145, opened 2017-05-11
Treat \n and \r same as other escaped characters

upsuper:
b'\n' and b'\r' should be serialized to '\\a ' and '\\d ' respectively, rather than '\\A ' and '\\D '. Simply let them be treated the same way as other escaped character should be fine.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-cssparser/145)


SimonSapin:
👍 

@bors-servo r+

PR #154, opened 2017-06-03
Parse nth-child(-/*comment*/n+b)

### LINKED ISSUE #153: :nth-child(-/**/n+2) should parse
According to [the spec](https://drafts.csswg.org/css-syntax-3/#the-anb-type), spaces are not allowed between `+n` and `-n` in `nth-child`, however it says nothing about comments. Chrome allows this, and with https://bugzilla.mozilla.org/show_bug.cgi?id=1364009, Firefox will too. We should as well.

cc @SimonSapin @dbaron

Manishearth:
fixes #153

r? @emilio

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-cssparser/154)


src/nth.rs
```diff
@@ -37,12 +38,17 @@ pub fn parse_nth(input: &mut Parser) -> Result<(i32, i32), ()> {
                 }
             }
         }
-        Token::Delim('+') => match try!(input.next_including_whitespace()) {
-            Token::Ident(value) => {
-                match_ignore_ascii_case! { &value,
-                    "n" => parse_b(input, 1),
-                    "n-" => parse_signless_b(input, 1, -1),
-                    _ => Ok((1, try!(parse_n_dash_digits(&*value))))
+        Token::Delim('+')| Token::Delim('-') => match try!(input.next_including_whitespace()) {
```
emilio: nit: spacing.

But I'd write it as:

```rust
Token::Delim(delim) => {
    let val = match delim {
        '+' => 1,
        '-' => -1,
        _ => return Err(()),
    };
    match try!(input.next_including_whitespace()) {
        // ...
    }
}
```

That way you avoid the extra variable at the top too.

Manishearth:
Yeah, so the spec is weird here. It explicitly includes a specification for all kinds of tokens, including tokens like `3n-` (which is a dimension token). Based on these, neither whitespace nor comments are allowed within any token, as is usual. The `3n` is never parsed as two tokens, so it never has whitespace. This was the main point of the patch.

_On top of that_, the `-n` _is_ parsed as two tokens, _but_ does not allow whitespace, as is specced. Chrome takes this to mean "comments but not whitespace", and Firefox supported comments and whitespace already. So I changed it to be uniformly "comments but not whitespace" everywhere, since by default you can assume that the spec is ignoring comments all the time.

Yes, that probably could be clarified more in that commit message but it's long :) I will try.

In this cssparser PR, i don't really change much behavior, I just make -n behave the the same as +n, which should match up regardless of your interpretation of the rest.

emilio:
Let's wait until the discussion in #153 settles though.

Manishearth:
https://github.com/w3c/csswg-drafts/pull/1498

SimonSapin:
r=me if https://github.com/w3c/csswg-drafts/pull/1498 is accepted. (I’m officially co-editor of that spec so I could accept that change, but this section is tricky and Tab wrote it so I’d rather have him do it.)

tabatkins:
:-1:, this appears to be a total misreading of the Syntax spec.  More details in <https://github.com/w3c/csswg-drafts/pull/1498>

Manishearth:
This means that these tests only require Firefox fixes to pass now.

PR #159, opened 2017-06-15
Reduce the size of Token

SimonSapin:
Hopefully improving parsing performance: https://bugzilla.mozilla.org/show_bug.cgi?id=1347408

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-cssparser/159)

### DIFF
--- a/src/compact_cow_str.rs
+++ b/src/compact_cow_str.rs
@@ -105,7 +105,7 @@ impl<'a> Clone for CompactCowStr<'a> {
         if self.is_borrowed() {
             CompactCowStr { ..*self }
         } else {
-            Self::from(Box::from(&**self))
+            Self::from(String::from(&**self).into_boxed_str())
         }
     }
 }
@@ -142,7 +142,7 @@ impl<'a> From<CompactCowStr<'a>> for Cow<'a, str> {
             if is_borrowed {
                 Cow::Borrowed(&*raw)
             } else {
-                Cow::Owned(String::from(Box::from_raw(raw as *mut str)))
+                Cow::Owned(Box::from_raw(raw as *mut str).into_string())
             }
         }
     }
--- a/src/compact_cow_str.rs
+++ b/src/compact_cow_str.rs
@@ -1,0 +1,231 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+use std::borrow::{Borrow, Cow};
+use std::cmp;
+use std::fmt;
+use std::hash;
+use std::marker::PhantomData;
+use std::mem;
+use std::ops::Deref;
+use std::slice;
+use std::str;
+
+// All bits set except the highest
+const MAX_LEN: usize = !0 >> 1;
+
+// Only the highest bit
+const OWNED_TAG: usize = MAX_LEN + 1;
+
+/// Like `Cow<'a, str>`, but with smaller `std::mem::size_of`. (Two words instead of four.)
+pub struct CompactCowStr<'a> {
+    // `tagged_len` is a tag in its highest bit, and the string length in the rest of the bits.
+    //
+    // * If the tag is 1, the memory pointed to by `ptr` is owned
+    //   and the lifetime parameter is irrelevant.
+    //   `ptr` and `len` are the components of a `Box<str>`.
+    //
+    // * If the tag is 0, the memory is borrowed.
+    //   `ptr` and `len` are the components of a `&'a str`.
+
+    // FIXME: https://github.com/rust-lang/rust/issues/27730 use NonZero or Shared
+    ptr: *const u8,
+    tagged_len: usize,
+    phantom: PhantomData<&'a str>,
+}
+
+impl<'a> From<&'a str> for CompactCowStr<'a> {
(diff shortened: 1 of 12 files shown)


src/compact_cow_str.rs
```diff
@@ -0,0 +1,243 @@
...
+    #[inline]
+    pub fn as_str(&self) -> Option<&'a str> {
+        if self.is_borrowed() {
+            Some(unsafe { &*self.as_raw_str() })
+        } else {
+            None
+        }
+    }
+
+    /// Convert into `String`, re-using the memory allocation it was already owned.
```
emilio: nit: `if it was already owned`.

src/compact_cow_str.rs
```diff
@@ -0,0 +1,243 @@
...
+    }
+}
+
+impl<'a> Clone for CompactCowStr<'a> {
+    #[inline]
+    fn clone(&self) -> Self {
+        if self.is_borrowed() {
+            CompactCowStr { ..*self }
+        } else {
+            Self::from(String::from(&**self).into_boxed_str())
```
emilio: Why can't this create a borrowed version when cloning, given it has the same lifetime?

emilio: err, nevermind, I think I see why.

emilio:
Tests need an update, and feel free to wait for @mbrubeck to review too, in case he has any other comment :)

src/compact_cow_str.rs
```diff
@@ -0,0 +1,243 @@
...
+    ptr: *const u8,
+    tagged_len: usize,
+    phantom: PhantomData<&'a str>,
+}
+
+impl<'a> From<&'a str> for CompactCowStr<'a> {
+    #[inline]
+    fn from(s: &'a str) -> Self {
+        let len = s.len();
+        assert!(len <= MAX_LEN);
```
mbrubeck: At least in the current implementation of libstd, `String` and `Vec` in Rust are already restricted to a maximum capacity of `isize::MAX`, so this doesn't add any further restrictions on the length.

src/compact_cow_str.rs
```diff
@@ -0,0 +1,228 @@
...
+use std::mem;
+use std::ops::Deref;
+use std::slice;
+use std::str;
+
+// All bits set except the highest
+const MAX_LEN: usize = !0 >> 1;
+
+// Only the highest bit
+const TAG: usize = MAX_LEN + 1;
```
mbrubeck: Reading this code, I have remembering whether the tag bit is set for borrowed or owned strings.  Perhaps it should be renamed to `OWNED` or similar.

src/compact_cow_str.rs
```diff
@@ -0,0 +1,228 @@
...
+        CompactCowStr {
+            ptr: ptr,
+            tagged_len: len | TAG,
+            phantom: PhantomData,
+        }
+    }
+}
+
+impl<'a> CompactCowStr<'a> {
+    /// Wether this string refers to borrowed memory
```
mbrubeck: Typo: "Whether"

src/compact_cow_str.rs
```diff
@@ -0,0 +1,243 @@
...
+    // `tagged_len` is a tag in its highest bit, and the string length in the rest of the bits.
+    //
+    // * If the tag is 1, the memory pointed to by `ptr` is owned
+    //   and the lifetime parameter is irrelevant.
+    //   `ptr` and `len` are the components of a `Box<str>`.
+    //
+    // * If the tag is 0, the memory is borrowed.
+    //   `ptr` and `len` are the components of a `&'a str`.
+
+    ptr: *const u8,
```
SimonSapin: Sure, done. Although I think we never use this in `Option` or other places where `NonZero` would make a difference.

src/compact_cow_str.rs
SimonSapin: I don’t know if any bit on the pointer is guaranteed to be available. Maybe not high bits since it seems like a userland process can get up to 3 GB on 32-bit. Not low bits since a subslice of `&str` can start at any byte.

src/compact_cow_str.rs
SimonSapin: Done.

src/compact_cow_str.rs
SimonSapin: If `self` is borrowed, the new value will be borrowed too. If not, the lifetime parameter is irrelevant: it could be `'static`, and the owned memory does not necessarily live that long.

src/compact_cow_str.rs
SimonSapin: Done.

src/compact_cow_str.rs
SimonSapin: Done.

SimonSapin:
https://github.com/servo/servo/pull/17345#issuecomment-308865907 has lots of failures, let’s not land this until we figure that out.

SimonSapin:
@emilio, @mbrubeck, I got a green Servo try run, after fixing another double-free. (Missing `mem::forget` in a method that consumes `self`.) I’ll take your respective "Looks good" above as "r+" so that I can publish to make it easier to do a Gecko try run. Let me know if there is something else you’d like to fix and I’ll follow up.

@bors-servo r=emilio,mbrubeck

SimonSapin:
Rebase at the same base that I had been testing.

@bors-servo r=emilio,mbrubeck

PR #162, opened 2017-06-21
Serialize currentcolor as canonicalized form

canaltinova:
Serialize currentcolor as canonicalized form instead of currentColor. This is needed for servo/servo#15397
I'm not sure about version bump though. Is that a breaking change?

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-cssparser/162)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 
 name = "cssparser"
-version = "0.16.0"
+version = "0.17.0"
 authors = [ "Simon Sapin <simon.sapin@exyr.org>" ]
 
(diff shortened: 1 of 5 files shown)


Manishearth:
@bors-servo delegate+

r=me

yeah, do a minor version bump so that we can pull in the new version.

SimonSapin:
I think this does not need to be a breaking change. Since 0.17.0 is not yet on crates.io, we can still make change this to 0.16.1.

I’m not sure what @Manishearth meant to say, since "minor" could either mean "not breaking" or "second component, between major and micro components"

PR #168, opened 2017-07-04
Parser changes for Gecko integration

jdm:
This is a grab bag of changes that were important for getting Stylo tests passing.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-cssparser/168)

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -364,9 +364,9 @@ impl<'i: 't, 't> Parser<'i, 't> {
     #[inline]
     pub fn parse_entirely<F, T, E>(&mut self, parse: F) -> Result<T, ParseError<'i, E>>
     where F: FnOnce(&mut Parser<'i, 't>) -> Result<T, ParseError<'i, E>> {
-        let result = parse(self);
+        let result = parse(self)?;
         self.expect_exhausted()?;
-        result
+        Ok(result)
     }
 
     /// Parse a list of comma-separated values, all with the same syntax.
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -489,8 +489,7 @@ impl<'i: 't, 't> Parser<'i, 't> {
         match self.next()? {
             Token::UnquotedUrl(value) => Ok(value),
             Token::Function(ref name) if name.eq_ignore_ascii_case("url") => {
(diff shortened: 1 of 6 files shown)


src/tokenizer.rs
```diff
@@ -226,7 +226,7 @@ impl<'a> Tokenizer<'a> {
             input: input,
             position: 0,
             last_known_source_location: Cell::new((SourcePosition(0),
-                                                   SourceLocation { line: 1, column: 1 })),
+                                                   SourceLocation { line: 0, column: 0 })),
```
SimonSapin: `SourceLocation`’s doc-comments needs to up updated accordingly.

src/parser.rs
```diff
@@ -188,6 +188,11 @@ impl<'i: 't, 't> Parser<'i, 't> {
         }
     }
 
+    /// Return the current line that is being parsed.
```
SimonSapin: Newlines are not significant in CSS. Why is the current line relevant? Is this counting on authors not using minification, adding newlines between declarations, and not writing multi-line declarations?

jdm: Gecko's error messages include a `sourceLine` property which is expected to be the current line being parsed.

jdm:
Related Gecko work is in the second patch of https://bugzilla.mozilla.org/show_bug.cgi?id=1352669.

SimonSapin:
r+ with the doc-comment updated for line and columns numbers starting at 0 rather than 1.

I still think that "current line" is not the right thing to do, but I suppose adding it will help enable these tests and get more test coverage of other things more easily than changing the tests to do something else. So let’s add it.

SimonSapin:
@jdm It looks like you landed this without the doc-comment change I requested. Could you fix that? CC https://bugzilla.mozilla.org/show_bug.cgi?id=1380890

jdm:
Weird; I remember making that change.

jdm:
Wait, I did: https://github.com/servo/rust-cssparser/pull/168/files#diff-0bd70f3585cde6d2d2aff02435db9070L389

SimonSapin:
I misread the diff, sorry.

PR #170, opened 2017-07-06
actually exclude test files from packaging

### LINKED ISSUE #169: Don't package big-data-url.css and color3_hsl.json
They are files that are >100k and trigger the large file error when vendoring in mozilla-central. If we could avoid packaging them, that would be nice.

froydnj:
package.exclude takes glob syntax for determining what files to exclude
when excluding directories.

Fixes #169.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-cssparser/170)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,7 +12,7 @@ keywords = ["css", "syntax", "parser"]
 license = "MPL-2.0"
 build = "build.rs"
 
-exclude = ["src/css-parsing-tests"]
+exclude = ["src/css-parsing-tests/**", "src/big-data-url.css"]
 
 [dev-dependencies]
 rustc-serialize = "0.3"


froydnj:
This is untested, but I can see the test files vendored in Firefox's repo with the present setup, so the present setup is obviously doing the wrong thing.

froydnj:
Hm, actually I'm not sure if it should be `src/css-parsing-tests/**`, `src/css-parsing-tests/*`, or `src/css-parsing-tests/**/*`...

SimonSapin:
You can test by running `cargo package` and examining the tarball created somewhere under `target/`.

froydnj:
> You can test by running cargo package and examining the tarball created somewhere under target/.

Aha, thanks!  Running `cargo package` indicates that 1509db2 does indeed prevent the tests from being packaged, so I think this PR is good to go!

SimonSapin:
Could you also add `big-data-url.css` to the list? r=me with that. Thanks!

froydnj:
Done, thanks!

PR #171, opened 2017-07-18
Add a one-token cache

SimonSapin:
This makes the `Servo_DeclarationBlock_SetPropertyById` micro-benchmark being added in https://bugzilla.mozilla.org/show_bug.cgi?id=1344131 faster by ~10%.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-cssparser/171)


src/compact_cow_str.rs
```diff
@@ -9,106 +9,97 @@ use std::hash;
...
+    // "Instantiate" the generic function without calling it.
+    let _ = mem::transmute::<CompactCowStr<'a>, Option<CompactCowStr<'a>>>;
 }
 
 impl<'a> From<&'a str> for CompactCowStr<'a> {
     #[inline]
     fn from(s: &'a str) -> Self {
         let len = s.len();
-        assert!(len <= MAX_LEN);
+        assert!(len < usize::MAX);
```
emilio: I think this would be (slightly) clearer as `!= usize::MAX`.

src/compact_cow_str.rs
```diff
@@ -9,106 +9,97 @@ use std::hash;
...
-    pub fn len(&self) -> usize {
-        self.tagged_len & !OWNED_TAG
-    }
-
-    // Intentionally private since it is easy to use incorrectly.
     #[inline]
-    fn as_raw_str(&self) -> *const str {
-        unsafe {
-            str::from_utf8_unchecked(slice::from_raw_parts(self.ptr, self.len()))
+    fn unpack(&self) -> Result<&'a str, *const String> {
```
emilio: Could this please have some docs? I found the name not super-descriptive.

src/parser.rs
```diff
@@ -348,11 +356,49 @@ impl<'i: 't, 't> Parser<'i, 't> {
...
+            _ => {
+                token = self.input.tokenizer.next().map_err(|()| BasicParseError::EndOfInput)?;
+                match token {
+                    // Don’t cache whitespace or comment tokens.
+                    // A typical pattern is:
+                    //
+                    // ```
+                    // parser.try(|parser| {
+                    //     match parser.next() { … }
+                    // }).or_else(|| {
```
emilio: Won't this change behaviour if I do:

```rust
parser.try(|p| match parser.next_including_whitespace_and_comments() { .. })
    .or_else(|| match parser.next_including_whitespace_and_comments() { .. })
```

nox: AFAIU, the position won't be the same after the whitespace token and thus the cache won't be used.

emilio: Oh, right

SimonSapin:
> SetPropertyById faster by ~10%.

It also makes full stylesheet parsing slower, so let’s not land this just yet…

SimonSapin:
Alright, in the current state of this PR (with a bunch of changes to the Servo repository for which I’ll open a PR next), gecko benchmarks changes are:

* Full stylesheet parsing: unchanged (or within noise level)
* Parsing `(eCSSProperty_width, "10px")`: ~15% faster
* Parsing `(eCSSProperty_width, " 10px")` (note extra space): ~5% **slower**

The regression is unfortunate. An additional change (not caching whitespace tokens that are skipped by `Parser::next`) fixes it (makes the `" 10px"` bench ~15% faster than master) but also makes full stylesheet parsing ~20% slower, so I removed that change from the PR.

I’ll keep trying other things, but I’d still like landing this so that the Servo changes do not bitrot. r? @emilio

src/tokenizer.rs
```diff
@@ -245,6 +245,15 @@ impl<'a> Tokenizer<'a> {
     }
 
     #[inline]
+    pub fn see_function(&mut self, name: &str) {
+        if self.var_functions == SeenStatus::LookingForThem {
```
emilio: nit: I'd probably use early-return here.

emilio: and below.

src/color.rs
```diff
@@ -141,7 +141,7 @@ impl Color {
     ///
     /// FIXME(#2) Deprecated CSS2 System Colors are not supported yet.
     pub fn parse<'i, 't>(input: &mut Parser<'i, 't>) -> Result<Color, BasicParseError<'i>> {
-        let token = input.next()?;
+        let token = input.next()?.clone();
```
emilio: is this clone needed? couldn't you just use `match *token`? I guess maybe you can't parse the nested block below, is that it?

And, you need the clone below for the error?

src/color.rs
```diff
@@ -475,7 +471,7 @@ fn parse_rgb_components_rgb<'i, 't>(arguments: &mut Parser<'i, 't>) -> Result<(u
         }
         Token::Percentage { unit_value, .. } => {
             red = clamp_unit_f32(unit_value);
-            green = clamp_unit_f32(match arguments.next()? {
+            green = clamp_unit_f32(match arguments.next()?.clone() {
```
emilio: Hmm... This is kind of unfortunate, because it means that before we created a token, and now we need to create two copies of it... Does it happen too often? I wonder if that's why parsing a stylesheet becomes slower?

src/color.rs
SimonSapin: Cloning is needed so that `input` is not still borrowed by the time we call `input.parse_nested_block` in the `Token::Function` case below. This is language limitation, non-lexical lifetimes will hopefully make this clone unnecessary. I’ll add a comment.

You’re right though that this clone makes the other one unnecessary, I’ll remove that.

src/color.rs
SimonSapin: Parsing initially became slower when `Parser::next` would clone every single token in order to put in the cache. Returning `&Token` removed these clones. In a minority of cases like here, the callers had to do these clones themselves instead. Note though that `Token::clone` never allocates. (Refcounting is used when a token contains an allocation.)

I don’t know the reason for the more recent regression. I’ll investigate next week.

SimonSapin:
https://github.com/rust-lang/rust/pull/42565 adds `From<&str> for Rc<str>` but has not landed yet. Also, `Rc<str>` is a fat pointer, which would make this bit packing more tricky. (`Rc::<T>::into_raw` and `Rc::<T>::from_raw` are only available for `T: Sized`.)

SimonSapin:
Ideally we’d want a growable `RcString` that contains refcount, length, capacity, and data all in one allocation. I tried it, but the amount of rarely-exercised `unsafe` code this required didn’t seem worth it.

emilio:
Yeah, fair enough... Anyway, feel free to land with r=me when you feel comfortable with it.

SimonSapin:
Cool, thanks. Maybe https://github.com/servo/servo/pull/17820 should be ready too before we land this one. It has a green servo try run but still need review. Do you want to take it as well? It’s fairly mechanical.

PR #172, opened 2017-07-23
Derive Default for SourceLocation

canaltinova:
This will help me to reduce the code size of my work in servo.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-cssparser/172)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 
 name = "cssparser"
-version = "0.17.0"
+version = "0.17.1"
 authors = [ "Simon Sapin <simon.sapin@exyr.org>" ]
 
 description = "Rust implementation of CSS Syntax Level 3"
--- a/src/tokenizer.rs
+++ b/src/tokenizer.rs
@@ -395,7 +395,7 @@ pub struct SourcePosition(usize);
 
 
 /// The line and column number for a given position within the input.
-#[derive(PartialEq, Eq, Debug, Clone, Copy)]
+#[derive(PartialEq, Eq, Debug, Clone, Copy, Default)]
 pub struct SourceLocation {
     /// The line number, starting at 0 for the first line.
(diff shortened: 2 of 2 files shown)


SimonSapin:
Where would you use this? `(0, 0)` marks the start of the input. Is this desired? Wouldn’t an obviously invalid location like `(u32::MAX, u32::MAX)` be preferable? Does it need to be the `Default` trait rather than some constructor?

emilio:
Yeah, I don't think `(0, 0)` it's a good default... Why is this needed?

canaltinova:
I'm implementing `@font-feature-values` rule and I need this to be able to derive Default for `FontFeatureValuesRule` struct. That means empty `FontFeatureValuesRule`, I think we don't need to set MAX value or other values for this because these values will be overwritten at some point. Also `FontFaceRuleData` has [something similar](https://dxr.mozilla.org/servo/rev/d8b3f9bc5c2837ca55ce474be91bf0f5e111ff71/components/style/font_face.rs#237). This method can also be changed with Default derive after this.
I'm okay to change this with a constructor named `zero` or something. But that will make deriving Default for `FontFeatureValuesRule` or `FontFaceRuleData` not possible.

emilio:
I don't think `empty()` should be implemented using `Default`, conceptually... But no strong opinion I guess. This seems somewhat easy to misuse, that's all.

So I'm fine with whatever @SimonSapin decides.

SimonSapin:
That `empty` constructor should take a `SourceLocation` as a parameter.

Using `Default` would allow writing shorter code, but I think in this case it is wrong conceptually. Please write the longer code, sorry :) (Like for `FontFaceRuleData::empty`, the repetitive parts can probably be generated by a macro.)

canaltinova:
Oh okay, I just wanted to shorten the code but if you think it can be misused, we can close this.

PR #173, opened 2017-07-24
Serialize number to no more than necessary precision

upsuper:
I'm not sure whether it can be simplified... and I probably need some more idea for how should I test it... But things look fine so far.

This should probably goes to an independent crate at some point.

It should fix servo/servo#17205.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-cssparser/173)

### DIFF
--- a/src/serializer.rs
+++ b/src/serializer.rs
@@ -2,8 +2,10 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+use dtoa_short::{self, Notation};
 use std::ascii::AsciiExt;
 use std::fmt::{self, Write};
+use std::str;
 
 use super::Token;
 
@@ -41,7 +43,6 @@ pub trait ToCss {
     }
 }
 
-
 #[inline]
 fn write_numeric<W>(value: f32, int_value: Option<i32>, has_sign: bool, dest: &mut W)
                     -> fmt::Result where W: fmt::Write {
@@ -50,15 +51,18 @@ fn write_numeric<W>(value: f32, int_value: Option<i32>, has_sign: bool, dest: &m
         dest.write_str("+")?;
     }
 
-    if value == 0.0 && value.is_sign_negative() {
+    let notation = if value == 0.0 && value.is_sign_negative() {
         // Negative zero. Work around #20596.
-        dest.write_str("-0")?
+        dest.write_str("-0")?;
+        Notation { decimal_point: false, scientific: false }
     } else {
-        write!(dest, "{}", value)?
-    }
+        dtoa_short::write(dest, value)?
+    };
 
     if int_value.is_none() && value.fract() == 0. {
-        dest.write_str(".0")?;
+        if !notation.decimal_point && !notation.scientific {
+            dest.write_str(".0")?;
+        }
     }
     Ok(())
 }
@@ -267,7 +271,7 @@ impl<'a, W> fmt::Write for CssStringWriter<'a, W> where W: fmt::Write {
 }
 
 
-macro_rules! impl_tocss_for_number {
+macro_rules! impl_tocss_for_int {
     ($T: ty) => {
         impl<'a> ToCss for $T {
             fn to_css<W>(&self, dest: &mut W) -> fmt::Result where W: fmt::Write {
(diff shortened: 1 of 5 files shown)


upsuper:
cc @nox @bzbarsky @Manishearth

SimonSapin:
After experimenting a bit, I believe that the standard library’s default float formatting already uses the shortest representation that parses to the same float. The issue in https://github.com/servo/servo/issues/17205 is specifically with percentages. I can reproduce it with the given test case, but not when changing `%` to `px`. This is because we divide the value of percentage tokens by `100.0` during parsing, and multiply it by `100.0` during serialization.

`"15%"` does not serialize as `"15%"` because  `(15_f32 / 100.) * 100.` is not equal to `15_f32`.

Float formatting can be very subtle, and I’m not at all familiar with the details. Responsibility for maintaining code in this repository is de-facto mine, so I’d much rather find another solution than adding a bunch of float formatting code. What do you think of not dividing percentage token value in the parser and only doing so in an accessor for layout code?

bzbarsky:
> I believe that the standard library’s default float formatting already uses the shortest representation that parses to the same float.

That's true, but the "same float" is not the important thing here.  I'll follow up in servo/servo#17205, though I'll be largely repeating things I already said there....

SimonSapin:
Alright, it sounds like we do want to round to 6 decimal places (from the first significant digit, not from the fraction separator). Still, I’d feel better if that code were not in `cssparser`. @upsuper, would you agree to maintain it in a separate crate?

upsuper:
I'm okay with maintaining it in a separate crate, although I don't have a good sense of how should I name that crate.

upsuper:
I created a new crate [`dtoa-short`](https://github.com/upsuper/dtoa-short) for that purpose. It would be appreciated if you can also review the code in that crate.

src/serializer.rs
```diff
@@ -50,15 +52,20 @@ fn write_numeric<W>(value: f32, int_value: Option<i32>, has_sign: bool, dest: &m
...
-    if value == 0.0 && value.is_sign_negative() {
+    let result = if value == 0.0 && value.is_sign_negative() {
         // Negative zero. Work around #20596.
-        dest.write_str("-0")?
+        Cow::Borrowed("-0")
     } else {
-        write!(dest, "{}", value)?
-    }
+        let mut result = String::new();
+        dtoa_short::write(&mut result, value)?;
```
SimonSapin: Why can’t this write write to `dest` directly? Can we find some way to avoid allocating a `String`?

SimonSapin: Oh, I see `contains` below. Could something be added to `dtoa_short` so that it tells us that information?

SimonSapin:
I read through `dtoa-short`, the code looks ok. @upsuper, could you run `cargo owner -a github:servo:cargo-publish dtoa-short` to allow people in the servo team to publish a new version in case you’re on vacation or otherwise away?

upsuper:
I got `error: failed to add owners to crate dtoa-short: api errors: only members of github:servo:cargo-publish can add it as an owner`. Probably you need to add me into the team first? Or probably I can add you to the owner of the crate, and you can add servo team then.

src/serializer.rs
upsuper: In the initial commit here I added a "has_exp" out parameter, but that feels pretty broken as a public interface... Probably I can add a return value for `dtoa_short::write` to return the form of the number serialized.

SimonSapin: Unless you need to extend and existing value, out parameters are not idiomatic in Rust and can be replaced by returning a tuple. Or this case, replacing `Result<(), Error>` with `Result<bool, Error>`, or maybe a custom enum instead of `bool`.

upsuper:
Updated the patch with updated dtoa-short.

SimonSapin:
r=me

Presumably you’ll also want a version number bump to get this on crates.io?

upsuper:
Oh, hmm, yep.

Scotchester:
I want to call out [@bzbarsky's comment from servo/servo#17205](https://github.com/servo/servo/issues/17205#issuecomment-318707564):

> The basic question is what the thing is that we want to preserve.
> 
> The current behavior assumes that the floating point value is the thing we care about and that serializing should preserve it with as much fidelity as possible. But from the point of view of a web author the internal floating-point representation is not something they care about. Furthermore, it's commonly lossy from the web developer's point of view in all sorts of cases. That includes not just percentages, but various `calc()` cases, computed style cases, etc.
> 
> The other option, and what I think we should do, is to aim at preserving "the original string representation" when serializing. This is impossible in general, of course, because lots of strings map to the same float. But all 6-digit decimal strings can be encoded in distinct floats. So if we always serialize to a 6-digit string, we may produce something that parses to a different float value (or may not, depending on what processing happens in addition to the parsing), but we are more likely to produce something that round-trips the actual original input.

I ran into an issue today where I have some SCSS being run through Parcel – whose [Lightning CSS](https://github.com/parcel-bundler/lightningcss) transformer relies on rust-cssparser – and the calculated values in the compiled CSS were all being rounded to 6 digits. This resulted in some values that weren't exactly what I was expecting.

This project is well over my head, but the quoted comment from @bzbarsky resonates with me – doing this sort of rounding in some cases results in "lossy" conversions from Sass/SCSS to CSS. Is there anything that can be done?

PR #176, opened 2017-07-25
Fix another panic in bad-url token parsing

SimonSapin:
https://bugzilla.mozilla.org/show_bug.cgi?id=1383975

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-cssparser/176)


emilio:
@bors-servo r+

Whoops, so it wasn't the same after all.

PR #177, opened 2017-08-09
Count line-numbers during tokenization

SimonSapin:
Before this PR, the `source_location` and `current_source_location` iterated (again) through input bytes separately from tokenization in order to count newline characters and determine the line number of some piece of a stylesheet.

This PR makes this counting happen during tokenization instead, where we already have a pass looking at every bytes.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-cssparser/177)

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -6,7 +6,7 @@ use cow_rc_str::CowRcStr;
 use std::ops::Range;
 use std::ascii::AsciiExt;
 use std::ops::BitOr;
-use tokenizer::{self, Token, Tokenizer, SourceLocation};
+use tokenizer::{Token, Tokenizer, SourcePosition, SourceLocation};
 
 
 /// A capture of the internal state of a `Parser` (including the position within the input),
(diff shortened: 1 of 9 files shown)


SimonSapin:
Before this PR, making `source_location` them return a dummy value immediately made Stylo stylesheet parsing faster by ~5%.

I *think* this PR makes stylesheet parsing faster by ~1% or so, but it’s hard to tell because this is close to noise level. This is a bit disappointing (I was hoping for more), but I think it’s worth landing anyway.

r? @emilio

src/parser.rs
```diff
@@ -6,20 +6,40 @@ use cow_rc_str::CowRcStr;
...
 ///
 /// Can be used with the `Parser::reset` method to restore that state.
 /// Should only be used with the `Parser` instance it came from.
-#[derive(PartialEq, Eq, Debug, Clone, Copy)]
-pub struct SourcePosition {
-    position: tokenizer::SourcePosition,
-    at_start_of: Option<BlockType>,
+#[derive(Debug, Clone)]
+pub struct ParserState {
+    pub(crate) position: usize,
```
emilio: Oh, the fancyness is coming? :)

src/tokenizer.rs
```diff
@@ -649,12 +699,16 @@ fn consume_quoted_string<'a>(tokenizer: &mut Tokenizer<'a>, single_quote: bool)
...
+                        b'\n' | b'\x0C' => {
+                            tokenizer.advance(1);
+                            tokenizer.seen_newline(false);
+                        }
                         b'\r' => {
                             tokenizer.advance(1);
                             if tokenizer.next_byte() == Some(b'\n') {
                                 tokenizer.advance(1);
                             }
+                            tokenizer.seen_newline(false);
```
emilio: This is slightly confusing, but I guess it's nicer than call seen_newline with true before this... maybe deserves a comment?

src/tokenizer.rs
```diff
@@ -921,24 +975,56 @@ unsafe fn from_utf8_release_unchecked(string_bytes: Vec<u8>) -> String {
...
-        match_byte! { c,
-            b' ' | b'\t' | b'\n' | b'\r' | b'\x0C' => {},
+    let start_position = tokenizer.position;
+    let from_start = &tokenizer.input[tokenizer.position..];
+    let mut newlines = 0;
+    let mut last_newline = 0;
+    let mut found_printable_char = false;
+    let mut iter = from_start.bytes().enumerate();
+    loop {
+        let (offset, b) = if let Some(item) = iter.next() {
```
emilio: nit: I think this'd be slightly easier to read with `match`.

src/parser.rs
SimonSapin: Yes, this raises the minimum Rust version for cssparser to 1.18. https://bugzilla.mozilla.org/show_bug.cgi?id=1383311 already made it 1.19 for Firefox.

src/tokenizer.rs
SimonSapin: Added a comment.

src/tokenizer.rs
SimonSapin: Done.

PR #180, opened 2017-08-18
Have at-rule without block handled in two stages as well

upsuper:
This change makes `AtRuleParser::parse_prelude` always return a prelude if parsed successfully, so that at-rules which do not accept block can be handled in two phases just like those with block.

This is important because at-rules without block usually have side effects. If `parse_prelude` executes the side effect, but `parse_at_rule` later determined that the rule is actually invalid (e.g. because it is followed by a block), there is no way for impl of `AtRuleParser` to undo the side effect.

It provides the necessary parser side change for fixing [Gecko bug 1388911](https://bugzilla.mozilla.org/show_bug.cgi?id=1388911).

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-cssparser/180)

### DIFF
--- a/src/css-parsing-tests/one_rule.json
+++ b/src/css-parsing-tests/one_rule.json
@@ -7,13 +7,17 @@
 "@foo", ["at-rule", "foo", [], null],
 
 "@foo bar; \t/* comment */", ["at-rule", "foo", [" ", ["ident", "bar"]], null],
-" /**/ @foo bar{[(4", ["at-rule", "foo",
+" /**/ @foo-with-block bar{[(4", ["at-rule", "foo-with-block",
     [" ", ["ident", "bar"]],
     [["[]", ["()", ["number", "4", 4, "integer"]]]]
 ],
 
-"@foo { bar", ["at-rule", "foo", [" "], [" ", ["ident", "bar"]]],
-"@foo [ bar", ["at-rule", "foo", [" ", ["[]", " ", ["ident", "bar"]]], null],
+"@foo-with-block { bar", ["at-rule", "foo-with-block", [" "], [
+    " ", ["ident", "bar"]]
+],
+"@foo [ bar", ["at-rule", "foo",
+    [" ", ["[]", " ", ["ident", "bar"]]], null
+],
 
 " /**/ div > p { color: #aaa;  } /**/ ", ["qualified rule",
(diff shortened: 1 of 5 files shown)


upsuper:
Probably we shouldn't take this until we can figure out a solution for [servo/servo#18139 (comment)](https://github.com/servo/servo/pull/18139#issuecomment-323310541).

upsuper:
Since that has been solved, I think this is ready for review as well now. This is a breaking change. I have a local patch for Servo side.

upsuper:
Maybe it is still better to distinguish between type for `WithBlock` and `WithoutBlock` so that users don't need to list unnecessary branches for prelude? But the existence of `OptionalBlock` is a problem then :/

SimonSapin:
I added `OptionalBlock` because it seemed easy enough at the time, but I don’t think any real CSS construct would ever use it. I don’t mind removing it.

upsuper:
OK, I'll try to figure out it tomorrow.

upsuper:
We probably should upgrade Servo to use 0.19.3 and then merge this PR and handle the change in a separate Servo PR.

SimonSapin:
https://github.com/servo/servo/pull/18171 upgrades to 0.19.3 but we’re waiting on https://github.com/dtolnay/dtoa/pull/9 to land it, unless you’re ok with including the unnecessary PNG in mozilla-central.

upsuper:
@SimonSapin ping for review?

SimonSapin:
Looks good.

Do tests pass at the intermediate commit? Please squash if not.

Do you have a corresponding Servo PR? r=me when that’s ready.

---

Reviewed 5 of 5 files at r1, 3 of 3 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/rust-cssparser/180#-:-Kss793_NYQ9N91uLEf-:b-60rlik)*

upsuper:
It is supposed to pass at the intermediate commit. I'll try again before merging.

upsuper:
So I can confirm that tests pass at the intermediate commit.

SimonSapin:
@bors-servo r+

---

Reviewed 2 of 2 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/rust-cssparser/180#-:-Ksw3s-7TUQ9E7zuDB-Y:bkg9nno)*

PR #182, opened 2017-08-22
Fix an off-by-one error in CDC skipping

tromey:
skip_cdc_and_cdo had an off-by-one error when trying to skip the CDC.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-cssparser/182)


SimonSapin:
Oops, silly mistake. Good catch, thanks!

@bors-servo r+

PR #188, opened 2017-08-28
Update encoding_rs to 0.7.0

hsivonen:
Update to the latest encoding_rs. The changes along the way don't appear to affect this crate.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-cssparser/188)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ exclude = ["src/css-parsing-tests/**", "src/big-data-url.css"]
 [dev-dependencies]
 rustc-serialize = "0.3"
 difference = "1.0"
-encoding_rs = "0.5"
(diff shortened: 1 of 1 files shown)


SimonSapin:
encoding_rs is only used in this repo for tests (it’s in `dev-dependencies`) so it shouldn’t affect downstream users. But thanks anyway!

@bors-servo r+

SimonSapin:
Stylo’s code (affecting what’s vendored in gecko) using encoding_rs is here: https://github.com/servo/servo/blob/master/components/style/encoding_support.rs

SimonSapin:
Looks like I forgot to rename Servo’s `RustEncoding` struct to `EncodingRs` when switch implementations…

PR #189, opened 2017-08-29
Clarify some dead code

tromey:
next_token has this code:

            if is_ident_start(tokenizer) { IDHash(consume_name(tokenizer)) }
            else if !tokenizer.is_eof() && match tokenizer.next_byte_unchecked() {
                b'a'...b'z' | b'A'...b'Z' | b'0'...b'9' | b'-' | b'_' => true,
                b'\\' => !tokenizer.has_newline_at(1),
                _ => !b.is_ascii(),
            } { Hash(consume_name(tokenizer)) }

I noticed that the `_` case incorrectly checks whether `b` is ASCII,
rather than the current byte.

However, on further inspection, it turns out that this can't be run,
because all such cases would have been filtered out by is_ident_start.

This patch clarifies this code by removing the dead cases and adding a
comment.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-cssparser/189)


SimonSapin:
Good catch, thanks!

@bors-servo r+

PR #190, opened 2017-08-29
Avoid std::fmt in serialization code

SimonSapin:
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-cssparser/190)

### DIFF
--- a/src/serializer.rs
+++ b/src/serializer.rs
@@ -3,8 +3,10 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 use dtoa_short::{self, Notation};
+use itoa;
 use std::ascii::AsciiExt;
 use std::fmt::{self, Write};
+use std::io;
 use std::str;
 
 use super::Token;
@@ -258,7 +260,32 @@ macro_rules! impl_tocss_for_int {
     ($T: ty) => {
         impl<'a> ToCss for $T {
             fn to_css<W>(&self, dest: &mut W) -> fmt::Result where W: fmt::Write {
-                write!(dest, "{}", *self)
+                struct AssumeUtf8<W: fmt::Write>(W);
+
+                impl<W: fmt::Write> io::Write for AssumeUtf8<W> {
+                    #[inline]
+                    fn write_all(&mut self, buf: &[u8]) -> io::Result<()> {
+                        // Safety: itoa only emits ASCII
+                        self.0.write_str(unsafe { str::from_utf8_unchecked(buf) })
+                            .map_err(|_| io::ErrorKind::Other.into())
+                    }
+
+                    #[inline]
+                    fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
+                        self.write_all(buf)?;
+                        Ok(buf.len())
+                    }
+
+                    #[inline]
+                    fn flush(&mut self) -> io::Result<()> {
+                        Ok(())
+                    }
+                }
+
+                match itoa::write(AssumeUtf8(dest), *self) {
+                    Ok(_) => Ok(()),
+                    Err(_) => Err(fmt::Error)
+                }
             }
         }
     }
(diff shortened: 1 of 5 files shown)


src/serializer.rs
```diff
@@ -90,7 +75,7 @@ impl<'a> ToCss for Token<'a> {
                 serialize_unquoted_url(&**value, dest)?;
                 dest.write_str(")")?;
             },
-            Token::Delim(value) => write!(dest, "{}", value)?,
+            Token::Delim(value) => dest.write_str(value.encode_utf8(&mut [0_u8; 4]))?,
```
emilio: Can't this use `write_char`?

src/serializer.rs
```diff
@@ -275,7 +290,32 @@ macro_rules! impl_tocss_for_int {
     ($T: ty) => {
         impl<'a> ToCss for $T {
             fn to_css<W>(&self, dest: &mut W) -> fmt::Result where W: fmt::Write {
-                write!(dest, "{}", *self)
+                struct AssumeUtf8<W: fmt::Write>(W);
+
+                impl<W: fmt::Write> io::Write for AssumeUtf8<W> {
+                    #[inline]
+                    fn write_all(&mut self, buf: &[u8]) -> io::Result<()> {
+                        // Safety: itoa only emits ASCII
```
emilio: Can we assert it?

```rust
debug_assert!(buf.iter().all(|b| b.is_ascii()));
```

src/serializer.rs
```diff
@@ -143,6 +132,32 @@ impl<'a> ToCss for Token<'a> {
...
+    let b4;
+    let bytes = if high > 0 {
+        let low = ascii_byte & 0x0F;
+        b4 = [b'\\', to_hex_byte(high), to_hex_byte(low), b' '];
+        &b4[..]
+    } else {
+        b3 = [b'\\', to_hex_byte(ascii_byte), b' '];
+        &b3[..]
+    };
+    dest.write_str(unsafe { str::from_utf8_unchecked(&bytes) })
```
emilio: This could use `write_char`, I think, but not sure if worth it, this would be faster and it's trivially ascii so...

src/tests.rs
```diff
@@ -300,10 +300,10 @@ fn unquoted_url_escaping() {
     let serialized = token.to_css_string();
     assert_eq!(serialized, "\
         url(\
-            \\1 \\2 \\3 \\4 \\5 \\6 \\7 \\8 \\9 \\A \\B \\C \\D \\E \\F \\10 \
-            \\11 \\12 \\13 \\14 \\15 \\16 \\17 \\18 \\19 \\1A \\1B \\1C \\1D \\1E \\1F \\20 \
+            \\1 \\2 \\3 \\4 \\5 \\6 \\7 \\8 \\9 \\a \\b \\c \\d \\e \\f \\10 \
+            \\11 \\12 \\13 \\14 \\15 \\16 \\17 \\18 \\19 \\1a \\1b \\1c \\1d \\1e \\1f \\20 \
```
emilio: Maybe run this through try to ensure this doesn't break anything? (seems unlikely)

src/serializer.rs
SimonSapin: Not sure why I didn’t think of that… Done :)

src/serializer.rs
SimonSapin: When writing more than one byte/char it might be better to minimize the number of calls, but I haven’t measured.

src/serializer.rs
SimonSapin: Done. (`AsciiExt` is also implemented on `[u8]`, no need to iterate ourselves.)

src/tests.rs
SimonSapin: I’ll do a Gecko Try run before landing in Servo.

PR #191, opened 2017-08-30
Change documentation of UnquotedUrl

tromey:
UnquotedUrl is only used for url() without quotes, so update the
documentation to say so.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-cssparser/191)


SimonSapin:
Thanks!

@bors-servo r+

PR #192, opened 2017-08-31
Utf 16 columns

tromey:
This series changes columns numbers to be reported in units of UTF-16 characters.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-cssparser/192)

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -811,6 +811,7 @@ pub fn parse_until_after<'i: 't, 't, F, T, E>(parser: &mut Parser<'i, 't>,
     let next_byte = (parser.input.tokenizer).next_byte();
     if next_byte.is_some() && !parser.stop_before.contains(Delimiters::from_byte(next_byte)) {
         debug_assert!(delimiters.contains(Delimiters::from_byte(next_byte)));
+        // We know this byte is ASCII.
         (parser.input.tokenizer).advance(1);
         if next_byte == Some(b'{') {
             consume_until_end_of_block(BlockType::CurlyBracket, &mut parser.input.tokenizer);
--- a/src/tests.rs
+++ b/src/tests.rs
@@ -1049,3 +1049,53 @@ fn roundtrip_percentage_token() {
         }
     }
 }
+
+#[test]
+fn utf16_columns() {
+    // This particular test serves two purposes.  First, it checks
+    // that the column number computations are correct.  Second, it
+    // checks that tokenizer code paths correctly differentiate
+    // between the different UTF-8 encoding bytes.  In particular
+    // different leader bytes and continuation bytes are treated
+    // differently, so we make sure to include all lengths in the
+    // tests, using the string "QΡ✈🆒".  Also, remember that because
+    // the column is in units of UTF-16, the 4-byte sequence results
+    // in two columns.
+    let tests = vec![
+        ("", 0),
+        ("ascii", 5),
+        ("/*QΡ✈🆒*/", 9),
+        ("'QΡ✈🆒*'", 8),
+        ("\"\\\"'QΡ✈🆒*'", 11),
+        ("\\Q\\Ρ\\✈\\🆒", 9),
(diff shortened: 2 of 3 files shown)


SimonSapin:
Looks good! A couple details:

---

Reviewed 1 of 1 files at r1, 2 of 2 files at r2, 2 of 2 files at r3.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[src/tokenizer.rs, line 371 at r1](https://reviewable.io:443/reviews/servo/rust-cssparser/192#-Kst0apvQrC_nCT35Q34:-Kst0apvQrC_nCT35Q35:b-n26mtd) ([raw file](https://github.com/servo/rust-cssparser/blob/c7c93567ff0be88680402dbca68c1afeec75be39/src/tokenizer.rs#L371)):*
> ```Rust
>         debug_assert!(byte == b'\r' || byte == b'\n' || byte == b'\x0C');
>         self.position += 1;
>         if byte == b'\r' && self.next_byte() == Some(/* LF */ b'\n') {
> ```

Nit: this "LF" comment was to mirror the former `is_cr` argument, but now that we have `\r` instead, `\n` is consistent and the comment can be removed.

---

*[src/tokenizer.rs, line 705 at r2](https://reviewable.io:443/reviews/servo/rust-cssparser/192#-KsszPL_r4mb3PHphIvM:-KsszPLaqh7c1LTbNXYm:b-2dnbed) ([raw file](https://github.com/servo/rust-cssparser/blob/32d69b7cd19dafdab65976a11d9620ce5d565741/src/tokenizer.rs#L705)):*
> ```Rust
>             b'\xF0'...b'\xFF' => { tokenizer.consume_4byte_intro(); }
>             _ => {
>                 tokenizer.advance(1);
> ```

Please add a comment here and in 4 other places with similar code below to indicate that the remaining cases are ASCII or leading bytes.

---

*[src/tokenizer.rs, line 378 at r3](https://reviewable.io:443/reviews/servo/rust-cssparser/192#-Kst-GlBGtlsJPtOs0EJ:-Kst-GlBGtlsJPtOs0EK:b-aknm2l) ([raw file](https://github.com/servo/rust-cssparser/blob/4a42697902960ad025cff6de79c1555d27d68fb2/src/tokenizer.rs#L378)):*
> ```Rust
>         // This takes two UTF-16 characters to represent, so we
>         // actually have an undercount.
>         self.current_line_start_position = self.current_line_start_position.wrapping_sub(1);
> ```

This uses `wrapping_sub` in case the very first byte of the input starts a 4-bytes code point, right? If so, shouldn’t every other operation with `current_line_start_position` use wrapping arithmetic? (Or maybe replacing `+= 1` in `consume_continuation_byte` with `wrapping_add` is enough, since it will bring the position back over zero with input that is known to be well-formed UTF-8?) Either way, please add a test for that.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/rust-cssparser/192#-:-Kst0nFYNCA4EzKWbuu5:bjlfwvg)*

tromey:
> This uses wrapping_sub in case the very first byte of the input starts a 4-bytes code point, right? If so, shouldn’t every other operation with current_line_start_position use wrapping arithmetic? (Or maybe replacing += 1 in consume_continuation_byte with wrapping_add is enough, since it will bring the position back over zero with input that is known to be well-formed UTF-8?) Either way, please add a test for that.

Yes, nice catch.  I believe the places where it is incremented must use wrapping_add; however, there's no need to use wrapping operations in `current_source_location`, because that should not be able to observe an "underflow" value (due to the UTF-8 encoding observation you point out).

SimonSapin:
@bors-servo r+

---

Reviewed 1 of 2 files at r5, 2 of 2 files at r6.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/rust-cssparser/192#-:-Ksvgd2AdsVfvpwYlUcz:bkg9nno)*

PR #197, opened 2017-09-13
Preserve sourceURL comment in parser

tromey:
In addition to the sourceMappingURL comment, there is a second special
comment, "sourceURL", that can be used to set the "display name" of a
style sheet for developer tools.  This name is also used as the base
URL for the source-map URL resolution algorithm.  sourceURL is
described here:
https://blog.getfirebug.com/2009/08/11/give-your-eval-a-name-with-sourceurl/
The relevant Firefox bug is here:
https://bugzilla.mozilla.org/show_bug.cgi?id=880831

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-cssparser/197)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "cssparser"
-version = "0.21.0"
+version = "0.21.1"
 authors = [ "Simon Sapin <simon.sapin@exyr.org>" ]
 
 description = "Rust implementation of CSS Syntax Level 3"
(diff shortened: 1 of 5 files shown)


SimonSapin:
Looks good. I assume you’re want this on crates.io, so r=me with the version number incremented.

SimonSapin:
I saw by chance that you had updated this PR but did not receive an email notification for that. I’ve seen github notification for further pushes to PR but they don’t seem to always happen… I don’t know if there’s a specific situation that causes that.

Anyway, please also comment (or ping on IRC, or something) when you push.

@bors-servo r+

PR #198, opened 2017-09-18
Saturate if nth-child value is out of range

Manishearth:
Reusing our tokenizer because it already saturates.

There is no way to reliably get the error details from `"-12034".parse()`, since the error type is opaque and the description strings may change.

This fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1401016

A less risky but also less correct fix is to simply return an error when `.parse()` fails.

r? @SimonSapin

Don't land until I can get a try run; nth-child has weird tokenization rules that we need to respect here (which is why I'm creating a new tokenizer in the first place)

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-cssparser/198)

### DIFF
--- a/src/nth.rs
+++ b/src/nth.rs
@@ -4,7 +4,7 @@
 
 use std::ascii::AsciiExt;
 
-use super::{Token, Parser, BasicParseError};
+use super::{Token, Parser, ParserInput, BasicParseError};
 
 
 /// Parse the *An+B* notation, as found in the `:nth-child()` selector.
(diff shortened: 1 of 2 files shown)


Manishearth:
Try passes. r?

https://treeherder.mozilla.org/#/jobs?repo=try&revision=7678094ea91f1c156493a52fd9597baf3cd2a022

src/nth.rs
```diff
@@ -93,8 +93,23 @@ fn parse_n_dash_digits(string: &str) -> Result<i32, ()> {
...
+        Ok(parse_number_saturate(&string[1..]).unwrap())  // Include the minus sign
     } else {
         Err(())
     }
 }
+
+fn parse_number_saturate(string: &str) -> Result<i32, ()> {
+    let mut input = ParserInput::new(string);
+    let mut parser = Parser::new(&mut input);
+    let int = if let Ok(&Token::Number {int_value: Some(int), ..})
```
heycam: The structure here feels a little weird.  How about:

```rust
  let next = parser.next_including_whitespace_and_comments();
  if let Ok(&Token::Number {int_value: Some(int), ..}) = next {
    if !parser.is_exhausted() {
      return Ok(int);
    }
  }
  Err(())
```

unless you're keen on having the Ok return at the end of the function.  Either way, probably nicer to have a separate `let next = ...`.

Manishearth:
Can't check is_exhausted whilst the parser is mutably borrowed by the token

SimonSapin:
Lexical lifetimes strike again :)

PR #201, opened 2017-11-11
include LICENSE to sub-crates

ignatenkobrain:
Signed-off-by: Igor Gnatenko <i.gnatenko.brain@gmail.com>

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-cssparser/201)


ignatenkobrain:
@SimonSapin would be very cool if you would release new version of crates with this as it would simplify my life as Fedora packager

SimonSapin:
Hi @ignatenkobrain. Please also increment the version number in `Cargo.toml` for the crates you’d like republished on crates.io.

ignatenkobrain:
@SimonSapin done

SimonSapin:
Both are publish. Thanks!

ignatenkobrain:
@SimonSapin thanks! and I just realized that procedurial-masquerade has MIT/Apache-2.0, not the MPL-2.0 license. Is it intentional?

SimonSapin:
That seems ok. I probably picked the same license as Rust out of habit when creating new crates and didn’t remember that cssparser is MPL2. cssparser only happens to be its first user, and I kept it in the same repository to not bother creating yet another one.

ignatenkobrain:
@SimonSapin eh, so we need to add MIT and Apache-2.0 license to that subcrate instead of just symlink to MPL text. Could you do this please? Sorry for not noticing earlier

PR #202, opened 2017-11-11
procedural-masquerade is not MPL

SimonSapin:
CC https://github.com/servo/rust-cssparser/pull/201#issuecomment-343659915

This change is [Reviewable](https://reviewable.io/reviews/servo/rust-cssparser/202)


SimonSapin:
@ignatenkobrain, how does this look?

ignatenkobrain:
@SimonSapin looks awesome!

SimonSapin:
Published 0.1.5

SimonSapin:
Out of curiosity, what are you packaging that depends on this crate?

ignatenkobrain:
@SimonSapin extprim_literals <- extprim <- syntex_syntax

I know that syntex is dead, but latest "old" rustfmt still depends on it. And we don't package nightly compiler, only stable.

In any case, it is also dependency of cssparser which is used by servo, isn't it? I would like to package servo one day as well. Not sure if it's used by FF too (probably some of deps of stylo..)

PR #203, opened 2017-11-20
Remove Column token.

heycam:
It no longer exists in the css-syntax spec.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-cssparser/203)


SimonSapin:
This is a breaking API change, please make sure you increment the version number accordingly before publishing on crates.io. (Ideally that would have been part of the PR.)

heycam:
Thanks, I'll try to remember about version bumps next time.

PR #211, opened 2018-01-11
Update syn to 0.12 and quote to 0.4

Eijebong:
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-cssparser/211)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,8 +30,10 @@ serde = {version = "1.0", optional = true}
 smallvec = "0.6"
 
 [build-dependencies]
-syn = "0.11"
-quote = "0.3"
+syn = { version = "0.12", features = ["extra-traits", "fold"] }
+quote = "0.4.1"
+proc-macro2 = "0.2"
+
 [features]
 bench = []
 dummy_match_byte = []
--- a/build.rs
+++ b/build.rs
@@ -2,8 +2,11 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+#[macro_use]
 extern crate quote;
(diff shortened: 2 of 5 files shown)


SimonSapin:
In `macros/lib.rs`, could we convert between `proc_macro::TokenStream` and `proc_macro2::TokenStream` without going explicitly through `String`? That way enabling the `proc_macro2/nightly` cargo feature would preserve span information and produce better error message if something fails to compile.

---

Reviewed 5 of 5 files at r1.

---

*[build/match_byte.rs, line 20 at r1](https://reviewable.io:443/reviews/servo/rust-cssparser/211#-L2_bQVhwr1tK9yL8RsP:-L2_bQVhwr1tK9yL8RsQ:bcboq5o) ([raw file](https://github.com/servo/rust-cssparser/blob/c3ce5db99665368f1c45e97d9117142cd13c05d2/build/match_byte.rs#L20)):*
> ```Rust
>     let mut source = String::new();
>     File::open(from).unwrap().read_to_string(&mut source).unwrap();
>     let tts = syn::parse_file(&source).expect("Parsing rules.rs module");
> ```

Rename `tts` (token trees) to something like `ast`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/rust-cssparser/211#-:-L2__GhgFAY5IwCCtnAc:b-qct142)*

SimonSapin:
Never mind, I forgot that these conversions to/from strings are in `precedural-masquerade`, refactoring that would be more involved than I thought.

@bors-servo r+

---

Reviewed 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/rust-cssparser/211#-:-L2_kKMpYCNAZGfS0cqP:bhyhujo)*

PR #212, opened 2018-01-24
Update difference to 2.0

ignatenkobrain:
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-cssparser/212)


emilio:
This is only used for testing, so looks good, thanks!

I just tested with a change that failed tests and it looks pretty much the same on my machine.

Thank you!

@bors-servo

PR #223, opened 2018-06-05
Update syn, quote, proc-macro2 and bump versions

Eijebong:


SimonSapin:
@bors-servo r+

Thanks!

PR #229, opened 2018-10-16
Update syn and bump version

Eijebong:
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-cssparser/229)

### DIFF
--- a/build/match_byte.rs
+++ b/build/match_byte.rs
@@ -8,6 +8,7 @@ use std::io::{Read, Write};
 use std::path::Path;
 use syn;
 use syn::fold::Fold;
+use syn::parse::{Parse, ParseStream, Result};
 
 use proc_macro2::{Span, TokenStream};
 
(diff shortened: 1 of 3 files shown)


build/match_byte.rs
```diff
@@ -27,20 +29,18 @@ pub fn expand(from: &Path, to: &Path) {
 
 struct MatchByte {
     expr: syn::Expr,
-    arms: Vec<syn::Arm>,
+    _comma: Token![,],
+    arms: Punctuated<syn::Arm, Option<Token![,]>>,
```
dtolnay: The `Arm` parser will always eat a trailing comma if there is one. https://github.com/dtolnay/syn/blob/0.15.12/src/expr.rs#L2777-L2781. The comma will go in https://docs.rs/syn/0.15/syn/struct.Arm.html#structfield.comma.

I would recommend parsing these as Vec\<Arm\> as seen in https://github.com/dtolnay/syn/blob/0.15.12/src/expr.rs#L1898-L1901.

build/match_byte.rs
```diff
@@ -27,20 +29,18 @@ pub fn expand(from: &Path, to: &Path) {
 
 struct MatchByte {
     expr: syn::Expr,
-    arms: Vec<syn::Arm>,
+    _comma: Token![,],
```
dtolnay: I would leave this out of the struct if it won't be used.

SimonSapin:
https://crates.io/crates/cssparser/0.25.1

PR #230, opened 2018-11-05
tokenizer: Look for both env() and var() functions.

emilio:
Differential Revision: https://phabricator.services.mozilla.com/D9609

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-cssparser/230)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "cssparser"
-version = "0.24.1"
+version = "0.25.0"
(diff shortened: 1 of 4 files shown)


SimonSapin:
I have some objections to the way this features is specified to work, but they are for CSSWG rather than this repo. In the meantime, let’s not block Firefox compat with what’s already shipping elsewhere.

@bors-servo r+

emilio:
(since cam took a look upstream)

PR #232, opened 2018-11-05
Bump cssparser-macros version

Eijebong:
forgot to do when I bumped syn

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-cssparser/232)


SimonSapin:
https://crates.io/crates/cssparser-macros/0.3.5

PR #234, opened 2018-11-19
Rustfmt everything

mandreyel:
I was reading through the code and noticed that the formatting was different from what I understand to be the currently agreed upon convention. So I ran rustfmt on the entire repo. I hope that's welcome. Tests run as previously.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-cssparser/234)

### DIFF
--- a/procedural-masquerade/lib.rs
+++ b/procedural-masquerade/lib.rs
@@ -244,33 +244,33 @@ macro_rules! define_invoke_proc_macro {
         #[doc(hidden)]
         #[macro_export]
         macro_rules! $macro_name {
-                    ($proc_macro_name: ident ! $paren: tt) => {
-                        #[derive($proc_macro_name)]
-                        #[allow(unused)]
-                        enum ProceduralMasqueradeDummyType {
-                            // The magic happens here.
-                            //
-                            // We use an `enum` with an explicit discriminant
-                            // because that is the only case where a type definition
-                            // can contain a (const) expression.
-                            //
-                            // `(0, "foo").0` evalutes to 0, with the `"foo"` part ignored.
-                            //
-                            // By the time the `#[proc_macro_derive]` function
-                            // implementing `#[derive($proc_macro_name)]` is called,
-                            // `$paren` has already been replaced with the input of this inner macro,
-                            // but `stringify!` has not been expanded yet.
-                            //
-                            // This how arbitrary tokens can be inserted
-                            // in the input to the `#[proc_macro_derive]` function.
-                            //
-                            // Later, `stringify!(...)` is expanded into a string literal
-                            // which is then ignored.
-                            // Using `stringify!` enables passing arbitrary tokens
-                            // rather than only what can be parsed as a const expression.
-                            Input = (0, stringify! $paren ).0
+                            ($proc_macro_name: ident ! $paren: tt) => {
+                                #[derive($proc_macro_name)]
+                                #[allow(unused)]
+                                enum ProceduralMasqueradeDummyType {
+                                    // The magic happens here.
+                                    //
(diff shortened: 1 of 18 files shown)


mandreyel:
Initially I hadn't noticed that Servo had custom settings for the formatter, so I added the config file and rerun rustfmt. Probably needs squashing.

procedural-masquerade/lib.rs
```diff
@@ -227,33 +244,33 @@ macro_rules! define_invoke_proc_macro {
...
-                    // Later, `stringify!(...)` is expanded into a string literal
-                    // which is then ignored.
-                    // Using `stringify!` enables passing arbitrary tokens
-                    // rather than only what can be parsed as a const expression.
-                    Input = (0, stringify! $paren ).0
-                }
-            }
-        }
-    }
+                            ($proc_macro_name: ident ! $paren: tt) => {
```
mandreyel: Is the extreme indentation of this macro definition as expected?

src/serializer.rs
```diff
@@ -381,16 +428,27 @@ impl TokenSerializationType {
...
-            DelimHash | DelimMinus | Number => matches!(other.0,
-                Ident | Function | UrlOrBadUrl | DelimMinus | Number | Percentage | Dimension),
-            DelimAt => matches!(other.0,
-                Ident | Function | UrlOrBadUrl | DelimMinus),
+            Ident => matches!(
+                other.0,
+                Ident |
+                    Function |
+                    UrlOrBadUrl |
+                    DelimMinus |
```
mandreyel: This also seems a little strange to me.

procedural-masquerade/lib.rs
emilio: Yeah, this looks wrong. Maybe revert it and file a rustfmt issue?

src/from_bytes.rs
```diff
@@ -30,34 +29,42 @@ pub trait EncodingSupport {
...
-            Some(protocol_encoding) => return protocol_encoding
-        }
+            Some(protocol_encoding) => return protocol_encoding,
+        },
     }
     let prefix = b"@charset \"";
     if css.starts_with(prefix) {
         let rest = &css[prefix.len()..];
         match rest.iter().position(|&b| b == b'"') {
             None => (),
```
emilio: This could use `if let`.

src/parser.rs
```diff
@@ -345,9 +349,14 @@ impl<'i: 't, 't> Parser<'i, 't> {
...
         let start = self.state();
         let result = match self.next() {
-            Err(BasicParseError { kind: BasicParseErrorKind::EndOfInput, .. }) => Ok(()),
+            Err(BasicParseError {
+                kind: BasicParseErrorKind::EndOfInput,
+                ..
+            }) => Ok(()),
             Err(e) => unreachable!("Unexpected error encountered: {:?}", e),
-            Ok(t) => Err(start.source_location().new_basic_unexpected_token_error(t.clone())),
+            Ok(t) => Err(start
```
emilio: This does look a bit weird as well.

mandreyel: Hmm, that's weird indeed. Putting them back on a single line fits under the line width limit too, so I'm not sure why this was formatted. Maybe another bug?
I'll file an issue for the macro bug (couldn't reply to that one for some reason).

SimonSapin:
Thanks!

I’ve rebased and integrated this in https://github.com/servo/rust-cssparser/pull/245. I removed the `rustfmt.toml` file since it’s more important IMO to be able to run stable rustfmt (and get formatting consistent with running it on Nightly) than to have the same style as servo/servo.

PR #239, opened 2019-02-27
Mark derive with `$crate::`

upsuper:
There is a problem currently that for example I create a procedural macro `cstr`, if users use
```rust
use cstr::cstr;
let x = cstr!(...);
```
the compiler would complain
> error: cannot find derive macro `cstr_internal__build_bytes` in this scope

This is because the custom derive macro is not imported. So if users want to use this macro, they would need to either use the traditional `#[macro_use] extern crate cstr;` or `use cstr::*;`. Neither is optimal.

This PR adds `$crate::` to the derive so that users of procedural macro can just import the macro they need.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-cssparser/239)

### DIFF
--- a/procedural-masquerade/Cargo.toml
+++ b/procedural-masquerade/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "procedural-masquerade"
-version = "0.1.6"
+version = "0.1.7"
 authors = ["Simon Sapin <simon.sapin@exyr.org>"]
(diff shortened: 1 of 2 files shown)


nox:
That's a breaking change, IIRC that's not supported by older Rust versions. If we make a breaking change we should rather remove hacks and use a proper procedural macro directly etc.

upsuper:
Hmmm... CI fails with interesting errors :/

The support of this syntax is from 1.30, fwiw.

PR #240, opened 2019-03-06
README.md fix ast link

mb21:
The `ast.rs` file does not exist anymore... so I'm assuming that's what's supposed to be linked to now?

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-cssparser/240)


SimonSapin:
Thanks!

PR #252, opened 2019-07-06
Use MaybeUninit when available 

SimonSapin:
Fix deprecation warnings for `mem::uninitialized`. (They also affected other crates, since it’s used in a public macro.)

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-cssparser/252)


SimonSapin:
I learned about it recently by looking at hashbrown

PR #255, opened 2019-07-17
Switch from rustc-serialize to serde_json

### LINKED ISSUE #253: Replace rustc-serialize dev-dependency with serde_json
serde 1.0 has been around [since over two years](https://github.com/serde-rs/serde/releases/tag/v1.0.0) now, so I think one could think about removing the rustc-serialize dev-dependency used in tests.

est31:
Fixes #253

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-cssparser/255)


SimonSapin:
Looks good, thanks!

@bors-servo r+

PR #256, opened 2019-08-09
Implement std::error::Error for errors

### LINKED ISSUE #247: PasrseError and error Trait
Hi,

I was wondering why the ParseError struct does not implement the [std::error::Error](https://doc.rust-lang.org/stable/std/error/trait.Error.html) trait? I was hoping to wrap it in a custom error enum type and use the "?" operator when calling parse functions. From what I've tried so far has not allowed me to do so.

xfix:
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-cssparser/256)

Fixes #247.

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -4,6 +4,8 @@
 
 use cow_rc_str::CowRcStr;
 use smallvec::SmallVec;
+use std::error::Error;
+use std::fmt;
 use std::ops::BitOr;
 use std::ops::Range;
 use tokenizer::{SourceLocation, SourcePosition, Token, Tokenizer};
@@ -53,6 +55,18 @@ pub enum BasicParseErrorKind<'i> {
     QualifiedRuleInvalid,
 }
 
+impl<'i> BasicParseErrorKind<'i> {
+    fn description(&self) -> &'static str {
+        match self {
+            BasicParseErrorKind::UnexpectedToken(_) => "Unexpected token",
+            BasicParseErrorKind::EndOfInput => "End of input",
+            BasicParseErrorKind::AtRuleInvalid(_) => "Invalid @ rule",
+            BasicParseErrorKind::AtRuleBodyInvalid => "Invalid @ rule body",
+            BasicParseErrorKind::QualifiedRuleInvalid => "Invalid qualified rule",
+        }
+    }
+}
+
 /// The funamental parsing errors that can be triggered by built-in parsing routines.
 #[derive(Clone, Debug, PartialEq)]
 pub struct BasicParseError<'i> {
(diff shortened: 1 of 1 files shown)


src/parser.rs
```diff
@@ -62,6 +76,14 @@ pub struct BasicParseError<'i> {
     pub location: SourceLocation,
 }
 
+impl<'i> fmt::Display for BasicParseError<'i> {
+    fn fmt(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
+        write!(formatter, "{}", self.kind.description())
+    }
+}
+
+impl<'i> Error for BasicParseError<'i> {}
```
nox: Why did you not implement `description` here?

xfix: The documentation for `Error` says: "This method is soft-deprecated. Although using it won’t cause compilation warning, new code should use Display instead and new impls can omit it."

dutchmartin:
@SimonSapin @nox Could someone take another look at this PR?
If there are any issues with it, please let me know.
I am happy to fix them since I want to use this error feature in my own project.

src/parser.rs
nox: Please implement it anyway, calling `BasicParseErrorKind::description`.

src/parser.rs
```diff
@@ -62,6 +76,14 @@ pub struct BasicParseError<'i> {
     pub location: SourceLocation,
 }
 
+impl<'i> fmt::Display for BasicParseError<'i> {
+    fn fmt(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
+        write!(formatter, "{}", self.kind.description())
```
nox: This won't print enough information, i.e. we end up not printing the payloads of `UnexpectedToken(_)` and `AtRuleInvalid(_)`.

src/parser.rs
```diff
@@ -156,6 +178,17 @@ impl<'i, T> ParseError<'i, T> {
     }
 }
 
+impl<'i, T> fmt::Display for ParseError<'i, T> {
+    fn fmt(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
+        write!(formatter, "{}", match &self.kind {
+            ParseErrorKind::Basic(basic_kind) => basic_kind.description(),
+            ParseErrorKind::Custom(_) => "Custom error",
+        })
```
nox: Please put the match outside and avoid `write!`.

```rust
match &self.kind {
    ParseErrorKind::Basic(basic) => basic.fmt(formatter),
    ParseErrorKind::Custom(_) => "custom error".fmt(formatter),
}
```

src/parser.rs
```diff
@@ -156,6 +178,17 @@ impl<'i, T> ParseError<'i, T> {
     }
 }
 
+impl<'i, T> fmt::Display for ParseError<'i, T> {
+    fn fmt(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
+        write!(formatter, "{}", match &self.kind {
+            ParseErrorKind::Basic(basic_kind) => basic_kind.description(),
+            ParseErrorKind::Custom(_) => "Custom error",
```
nox: Why is the payload disregarded here?

src/parser.rs
```diff
@@ -156,6 +178,17 @@ impl<'i, T> ParseError<'i, T> {
...
+        })
+    }
+}
+
+impl<'i, T> Error for ParseError<'i, T> where T: fmt::Debug {}
```
nox: Implement `description` here.

PR #258, opened 2019-08-25
Consolidate logic in match

twilco:
This was previously not possible due to lexical lifetimes.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-cssparser/258)


twilco:
Ah, didn't realize this crate is targeting Rust 2015 + `1.31.0`.  Closing this.

Issue #254, opened 2019-07-17
Future compat warning


est31:
Currently, there is a future compat warning firing when building the codebase:

````
warning: cannot specify lifetime arguments explicitly if late bound lifetime parameters are present
   --> src/rules_and_declarations.rs:537:30
    |
537 |         parse_until_before::<'i, 't, _, _, _>(input, Delimiter::CurlyBracketBlock, |input| {
    |                              ^^
    | 
   ::: src/parser.rs:959:13
    |
959 |     parser: &mut Parser<'i, 't>,
    |             - the late bound lifetime parameter is introduced here
    |
    = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
    = note: for more information, see issue #42868 <https://github.com/rust-lang/rust/issues/42868>
````

SimonSapin:
Yes, I am unhappy about this warning.

The diagnostic is not clear at all, and even after much discussion in https://github.com/rust-lang/rust/issues/42868 and on IRC I’m not sure what it means. None of the possible fixes suggested by various people are accepted by the compiler for this code.

So we’re stuck with an obscure warning we can’t silence and that no-one knows how to fix.

lqd:
For documentation purposes, here's a summary of the situation:
- there are currently 4 instances of code designed to avoid rustc bug https://github.com/rust-lang/rust/issues/42508 about region inference and implied bounds in closures: [#1](https://github.com/servo/rust-cssparser/blob/master/src/rules_and_declarations.rs#L284-L292), [#2](https://github.com/servo/rust-cssparser/blob/master/src/rules_and_declarations.rs#L486-L489), [#3](https://github.com/servo/rust-cssparser/blob/master/src/rules_and_declarations.rs#L535-L539), [#4](https://github.com/servo/rust-cssparser/blob/master/src/rules_and_declarations.rs#L544-L547).
- these 4 occurrences, in turn, now trigger the `late_bound_lifetime_arguments` lint tracked in https://github.com/rust-lang/rust/issues/42868, because of the lifetime annotations to both early and late-bound lifetimes for the generic `parse_*` function calls.
- in #260 I was able to avoid both of these by lifting the closure argument to a variable which can be removed in the future, and mentioned this issue in FIXME comments for more information. (Incidentally, 2 scopes also had to be added to make it work on 1.31.0, but are not technically required with more recent stable releases: they could removed if the minimum rust version were to be raised, for example to 1.36.0)

[...]

PR #260, opened 2019-10-15
Work around warning from `late_bound_lifetime_arguments` lint

lqd:
The generic parsing functions require lifetime annotations to build because of https://github.com/rust-lang/rust/issues/42508. 

However, this creates a warning from the `late_bound_lifetime_arguments` compatibility lint, described in https://github.com/rust-lang/rust/issues/42868, because they have both early-bound and late-bound lifetime parameters.

This PR works around this warning, in a functional but sub-optimal way. The result is quite verbose, and I didn't manage to "convince" rustc to think all lifetimes were early-bound at the `fn` definitions, but manages to avoid triggering both https://github.com/rust-lang/rust/issues/42508 and the lint, by passing these closures via variables at each call-site.

Fixes https://github.com/servo/rust-cssparser/issues/254

### DIFF
--- a/src/rules_and_declarations.rs
+++ b/src/rules_and_declarations.rs
@@ -281,15 +281,12 @@ where
                     // Ident
                     let result = {
                         let parser = &mut self.parser;
-                        // FIXME: https://github.com/rust-lang/rust/issues/42508
-                        parse_until_after::<'i, 't, _, _, _>(
-                            self.input,
-                            Delimiter::Semicolon,
-                            |input| {
-                                input.expect_colon()?;
-                                parser.parse_value(name, input)
-                            },
-                        )
+                        // FIXME: https://github.com/servo/rust-cssparser/issues/254
+                        let callback = |input: &mut Parser<'i, '_>| {
+                            input.expect_colon()?;
+                            parser.parse_value(name, input)
+                        };
+                        parse_until_after(self.input, Delimiter::Semicolon, callback)
                     };
                     return Some(result.map_err(|e| (e, self.input.slice_from(start.position()))));
                 }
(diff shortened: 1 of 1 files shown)


emilio:
Is there any good reason to do this other than building warning-free?

I don't think we should adapt your code for this necessarily, unless there's a very good reason to do this. Doubly so if this makes the code uglier tbf...

lqd:
Yes building without the warning was the reason for the change, as it's not clear when this is going to become an error.

emilio:
As long as this is working around a rust bug I don't think they can reasonably turn it into an error...

lqd:
Ah but this PR is not working around a rustc bug, it is "fixing" rust-cssparser's workaround for the rustc bug and which generates a compatiblity lint warning.

(the code is not that ugly on current stable. I haven't tested older stables so I don't know which one fixed the scope requirement, I only added it because CI fails on 1.31. These, and the existing ones, eg in parser.rs IIRC, could be removed on newer versions)

SimonSapin:
The code comment is useful, but perhaps the whole explanation does not need to be repeated every time. Maybe move it to a github comment in https://github.com/servo/rust-cssparser/issues/254, and have the code link there?

With that change I feel the ugliness would not be *too* bad. I think warning-freedom is valuable, although in this case we aren’t actively doing a lot of work on this crate so warnings haven’t been much of a bother.

src/rules_and_declarations.rs
```diff
@@ -483,10 +493,26 @@ where
...
+    // More information is available on the lint's tracking issue
+    // https://github.com/rust-lang/rust/issues/42868.
+    //
+    // Using this `callback` variable is a workaround for the lint's warning.
+    let result;
+    {
+        // Introduce a new scope to limit the duration of the `parser` loan
+        let callback = |input: &mut Parser<'i, '_>| parser.parse_prelude(name, input);
+        result = parse_until_before(input, delimiters, callback);
+    }
```
SimonSapin: Does the warning happen without assignment outside of `let`, like this?

```rust
let result = {
    …
};
```

If not, please change it and similarly `let prelude;` below.

lqd: this also works indeed, done

PR #261, opened 2019-10-18
Update phf to 0.8

nox:


SimonSapin:
`getrandom` failed to build on 1.31, https://github.com/rust-random/getrandom#minimum-supported-rust-version lists 1.32

nox:
> `getrandom` failed to build on 1.31, https://github.com/rust-random/getrandom#minimum-supported-rust-version lists 1.32

I know, which is why I want the `MaybeUninit` patch to land because it already increases minimum Rust version.

PR #263, opened 2019-10-21
Remove use of procedural-masquerade

SimonSapin:
It’s unnecessary since Rust 1.30


nox:
Nice.

@bors-servo r+

PR #267, opened 2019-11-21
Bump smallvec to 1.0, release new minor version (0.27.2)

Manishearth:
r? @SimonSapin


nox:
Isn't smallvec in the public API? In which case that's a breaking change.

nox:
Never mind, it is not.

PR #271, opened 2020-04-20
Format code using 'cargo fmt'

Atul9:


SimonSapin:
I would be ambivalent about taking this in isolation, but since https://github.com/servo/rust-cssparser/pull/275 is also ready to go this looks good. Thanks!

@bors-servo r+

PR #277, opened 2020-12-03
Pass rule start to parser rather than just a location.

emilio:
This is:

 * More generic, as the state has more information like the source
   index.

 * Slightly cheaper to compute, as the source location requires math to
   compute, which the parser state doesn't

I need this to improve CSS sanitization in Gecko.

https://bugzilla.mozilla.org/show_bug.cgi?id=1680084

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "cssparser"
-version = "0.27.2"
+version = "0.28.0"
 authors = [ "Simon Sapin <simon.sapin@exyr.org>" ]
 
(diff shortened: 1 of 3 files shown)


heycam (approved):

heycam:
Should we have a unit test for the fixed at-rule location?

emilio:
We have one integration test in Gecko for it and no existing tests for `SourceLocation` stuff, so it's not really trivial. I looked into adding one and it requires a bunch of boilerplate so not super-exciting...

@bors-servo r=heycam

PR #278, opened 2020-12-13
Remove procedural-masquerade.

emilio:
I don't think this is used anymore, so is there any point in keeping the
code around?


SimonSapin:
Although it isn’t used in cssparser anymore it is still used elsewhere: https://crates.io/crates/procedural-masquerade/reverse_dependencies. We could encourage those crate authors to move to plain `#[proc_macro]` if they require Rust 1.30+ (since procedural-masquerade is only useful on older compilers).

As long as it’s used having a source repository is occasionally useful to make fixes like https://github.com/servo/rust-cssparser/pull/274 / https://github.com/rust-lang/rust/pull/73345#issuecomment-646895843

emilio:
@SimonSapin should we move it to its own repo then? It seems a bit weird to have it in cssparser nowadays.

But anyhow, I'm ok closing this I guess, that's fair :)

PR #284, opened 2021-03-11
Add fuzzing support.

emilio:


SimonSapin:
Hey Emilio. I’m sorry but I don’t have much bandwidth available for Servo and related projects these days. If you feel that `cssparser` needs more reviewers I’ll trust you to designate them. I can then help with GitHub/bors/crates.io permissions.

emilio:
Sure, sounds good, thanks Simon :)

@BorisChiou would you be able to take a look at this?

BorisChiou:
Should be ok. I will take a look later today

fuzz/fuzz_targets/cssparser.rs
```diff
@@ -0,0 +1,96 @@
...
+    serialization
+}
+
+fn do_parse_and_serialize<'i>(
+    input: &mut Parser<'i, '_>,
+    preserving_comments: bool,
+    mut previous_token_type: TokenSerializationType,
+    serialization: &mut String,
+    indent_level: usize,
+) -> Result<(), ParseError<'i, ()>> {
```
BorisChiou, line 30: Just curious about the review type. Why not just return something like `Result<serialization, xxx>`?

fuzz/fuzz_targets/cssparser.rs
```diff
@@ -0,0 +1,96 @@
...
+            }
+            println!("{:?}", token);
+        }
+        if token.is_parse_error() {
+            let token = token.clone();
+            return Err(input.new_unexpected_token_error(token))
+        }
+        let token_type = token.serialization_type();
+        if previous_token_type.needs_separator_when_before(token_type) {
+            serialization.push_str("/**/");
```
BorisChiou, line 53: The spec says:
> If the tokenizer preserves comments, the preserved comment should be used; otherwise, an empty comment (/**/) must be inserted.
> https://drafts.csswg.org/css-syntax/#serialization

So we always insert an empty comment here because we don't support `preserving_comments` now. Ya, should be fine.

BorisChiou (approved):
So this fuzzing support looks like we just parse the input and serialize it as a better string? Not sure about the use case because there is no example for now. However, the overall logic looks fine to me.

jrmuizel:
Can this land now?

PR #287, opened 2021-09-01
Tweak at-rule parsing API to allow a single at-rule to be block or blockless.

emilio:
This is needed for @layer, which can be block or blockless.


BorisChiou (approved):
The overall looks good to me. Based on the usage of `@layer` rule, this update makes sense to me.

emilio:
Ah, I guess travis is no more. We'll have to update this to github actions sooner rather than later.

PR #289, opened 2021-09-25
Replace deprecated std constants by associated constants.

FelixRilling:
This PR aims to fix a few minor deprecation warnings.
See https://blog.rust-lang.org/2020/04/23/Rust-1.43.0.html#library-changes and https://doc.rust-lang.org/stable/std/usize/index.html for details.

### DIFF
--- a/src/cow_rc_str.rs
+++ b/src/cow_rc_str.rs
@@ -12,7 +12,6 @@ use std::ops::Deref;
 use std::rc::Rc;
 use std::slice;
 use std::str;
-use std::usize;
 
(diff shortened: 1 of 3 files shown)


jdm:
@bors-servo r+
Thanks!

jdm:
Looks like this has implications for the libraries minimum supported rust version.

FelixRilling:
Ah, good catch - I did not check that!

I guess merging this will have to wait until the min. rust version is 1.43 or higher then.

emilio:
@bors-servo r+

 * Should be good to go now.

PR #292, opened 2021-11-25
Implement support for hwb() color syntax

devongovett:
This implements support for the `hwb()` color syntax specified in the [CSS Color Module Level 4](https://drafts.csswg.org/css-color-4/#the-hwb-notation) spec. HWB colors are converted to RGB during parsing, just like HSL. Tests are generated using a script based on the [Web Platform Tests](https://github.com/web-platform-tests/wpt/blob/master/css/css-color/color-resolving-hwb.html).


emilio (approved):
Looks great to me. It'd be great to get some more code sharing between hsl and hwb, but I can land that in a follow-up.

emilio:
https://github.com/servo/rust-cssparser/pull/293 is the follow-up. Thanks again!

PR #293, opened 2021-11-26
Share more code between hsl and hwb parsing.

emilio:


GPHemsley:
I'm not sure this was the best decision.

`hsl()` and `hwb()` are defined independently of each other in the spec. They already have different support for legacy color syntax (i.e. commas) in css-colors-4, and their definitions could further diverge in the future (e.g. relative colors in css-colors-5).

In addition, in combining the code like this, you lose the correlation between the variable names and the spec, which makes the code harder to read.

I think you should revert these changes and take a different approach:
* The spec explicitly defines the `hue` processing as being identical between HSL and HWB (for now, at least), so that could still be a candidate for factoring out.
* And HWB doesn't support comma syntax, so that code can between removed from the corresponding function.

PR #295, opened 2021-12-24
color: Don't allow commas in hwb() function.

emilio:
It's not allowed per spec, though perhaps it should?

I filed https://github.com/w3c/csswg-drafts/issues/6911


emilio:
@borischiou / @hiikezoe / @upsuper if any of you can sanity-check this it'd be great, though it's quite trivial.

upsuper (commented):
LGTM

PR #296, opened 2022-01-25
Update itoa to 1.0

DBLouis:


emilio:
Thanks! Do you have some perf numbers for this by any chance? It seems this will introduce an extra buffer copy (to the stack though, so maybe fine).

It'd be good to avoid it somehow, but maybe it's just faster, or old itoa also did this.

DBLouis:
It seems that `itoa::write` was also using a buffer:
https://docs.rs/itoa/0.4.8/src/itoa/lib.rs.html#84-91

emilio:
Sorry for missing this! I merged this change in #302 and kept you as author.

PR #300, opened 2022-04-28
Update phf to 0.10

wusyong:
phf 0.8 still uses rand_core version <0.6.2 which suffer from [CVE-2021-27378](https://nvd.nist.gov/vuln/detail/CVE-2021-27378).
Every crate depends on it will be warned by cargo audit.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -16,7 +16,7 @@ jobs:
           - nightly
           - beta
           - stable
-          - 1.36.0
+          - 1.40.0
         features:
           -
           - --features dummy_match_byte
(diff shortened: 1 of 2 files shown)


CYBAI:
Should we make a patch version bump of cssparser for this phf upgrade 🤔 ?

wusyong:
It would great if that's so. servo-selector also suffer from it.

emilio:
Yeah feel free to bump the minor version as well before landing? r=me either way

emilio (approved):

emilio:
It seems this needs to update the minimum rust version, see the CI jobs. I think that's probably fine, 1.36 is kinda ancient.

wusyong:
Do you want to remove them (1.36) or set to the current one?

emilio:
@bors-servo r+

Thanks!

PR #304, opened 2022-05-21
color: Expose hsl/hwb to rgb functions.

emilio:
I'm going to use this for color-mix() support.


emilio:
(trivial)

emilio:
Homu is not cooperating, but CI is green.

PR #306, opened 2022-08-06
Include 0.11 in the range of compatible phf versions.

adamreichold:


adamreichold:
Will have to `cargo update --precise` the phf dependency for the MSRV run...

adamreichold:
> Will have to `cargo update --precise` the phf dependency for the MSRV run...

The step to downgrade the phf version for the MSRV job is a bit brittle because it references the current latest version of that crate. This can be avoided by using a bit of scripting like [here](https://github.com/PyO3/rust-numpy/blob/main/.github/workflows/ci.yml#L134) or by checking in a separate `Cargo.lock` which can be copied in for the MSRV job if the maintainers prefer any of these slightly more involved solutions.

emilio (approved):

adamreichold:
I am sorry for making this complicated by messing up the quoting rules for GitHub expressions. 🤦🏽

emilio:
@bors-servo r+

No problem, thanks!

PR #308, opened 2022-10-08
feat(color): Refactor existing color parsing to better align with spec

GPHemsley:
Add public structs `Hue` and `AlphaValue`.

Add methods `parse_hue()` and `parse_alpha_value()` to public trait `ColorComponentParser`.

Add or update documentation to include links to relevant sections of css-color-4 draft.

----

Following on from my comment on #293, I've gone ahead and refactored the color parsing code to better align with the text of the draft css-color-4 spec, which also allows for easier extension to support the multitude of new functionality that is being specified. (I already have a working implementation of the `none` keyword nearing completion based on these changes, for example.)

I don't know the policy on removing public features, so I've left in the `NumberOrPercentage` and `AngleOrNumber` enums and the `parse_angle_or_number()` and `parse_number_or_percentage()` methods of `ColorComponentParser`, even though I think they are now redundant. I'd be happy to remove them in a later PR, if that's deemed appropriate.

For reference, the latest draft of css-color-4 is available here:
https://w3c.github.io/csswg-drafts/css-color-4/

### DIFF
--- a/src/color.rs
+++ b/src/color.rs
@@ -11,6 +11,7 @@ use super::{BasicParseError, ParseError, Parser, ToCss, Token};
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
 
 /// A color with red, green, blue, and alpha components, in a byte each.
+/// https://w3c.github.io/csswg-drafts/css-color-4/#resolving-sRGB-values
 #[derive(Clone, Copy, PartialEq, Debug)]
 #[repr(C)]
 pub struct RGBA {
@@ -130,6 +131,7 @@ impl ToCss for RGBA {
 }
 
 /// A <color> value.
+/// https://w3c.github.io/csswg-drafts/css-color-4/#color-syntax
 #[derive(Clone, Copy, PartialEq, Debug)]
 pub enum Color {
     /// The 'currentcolor' keyword
@@ -165,15 +167,6 @@ pub enum NumberOrPercentage {
     },
 }
 
-impl NumberOrPercentage {
-    fn unit_value(&self) -> f32 {
-        match *self {
-            NumberOrPercentage::Number { value } => value,
-            NumberOrPercentage::Percentage { unit_value } => unit_value,
-        }
-    }
-}
-
 /// Either an angle or a number.
 pub enum AngleOrNumber {
     /// `<number>`.
@@ -188,13 +181,16 @@ pub enum AngleOrNumber {
     },
 }
 
-impl AngleOrNumber {
-    fn degrees(&self) -> f32 {
-        match *self {
-            AngleOrNumber::Number { value } => value,
-            AngleOrNumber::Angle { degrees } => degrees,
-        }
-    }
+/// https://w3c.github.io/csswg-drafts/css-color-4/#hue-syntax
+pub struct Hue {
+    /// The value as a number of degrees.
+    degrees: f32,
+}
+
+/// https://w3c.github.io/csswg-drafts/css-color-4/#alpha-syntax
+pub struct AlphaValue {
+    /// The value as a number in the range of 0 to 1.
+    number: f32,
 }
 
 /// A trait that can be used to hook into how `cssparser` parses color
@@ -259,6 +255,48 @@ pub trait ColorComponentParser<'i> {
         Ok(match *input.next()? {
             Token::Number { value, .. } => NumberOrPercentage::Number { value },
             Token::Percentage { unit_value, .. } => NumberOrPercentage::Percentage { unit_value },
+            ref t => return Err(location.new_unexpected_token_error(t.clone())),
+        })
+    }
+
+    /// Parse a `<hue>` value.
+    /// https://w3c.github.io/csswg-drafts/css-color-4/#typedef-hue
+    fn parse_hue<'t>(
+        &self,
+        input: &mut Parser<'i, 't>,
+    ) -> Result<Hue, ParseError<'i, Self::Error>> {
+        let location = input.current_source_location();
+        Ok(match *input.next()? {
+            Token::Number { value: degrees, .. } => Hue { degrees },
+            Token::Dimension {
+                value: v, ref unit, ..
+            } => {
+                let degrees = match_ignore_ascii_case! { &*unit,
+                    "deg" => v,
+                    "grad" => (v / 400.) * 360.,
+                    "rad" => (v / (2. * PI)) * 360.,
+                    "turn" => v * 360.,
+                    _ => return Err(location.new_unexpected_token_error(Token::Ident(unit.clone()))),
+                };
+
+                Hue { degrees }
+            }
+            ref t => return Err(location.new_unexpected_token_error(t.clone())),
+        })
+    }
+
+    /// Parse a `<alpha-value>` value.
+    /// https://w3c.github.io/csswg-drafts/css-color-4/#typedef-alpha-value
+    fn parse_alpha_value<'t>(
+        &self,
+        input: &mut Parser<'i, 't>,
+    ) -> Result<AlphaValue, ParseError<'i, Self::Error>> {
+        let location = input.current_source_location();
+        Ok(match *input.next()? {
+            Token::Number { value: number, .. } => AlphaValue { number },
+            Token::Percentage {
+                unit_value: number, ..
+            } => AlphaValue { number },
             ref t => return Err(location.new_unexpected_token_error(t.clone())),
         })
     }
@@ -350,6 +388,8 @@ fn rgba(red: u8, green: u8, blue: u8, alpha: u8) -> Color {
 /// Matching is case-insensitive in the ASCII range.
 /// CSS escaping (if relevant) should be resolved before calling this function.
 /// (For example, the value of an `Ident` token is fine.)
+///
+/// https://w3c.github.io/csswg-drafts/css-color-4/#color-keywords
 #[inline]
 pub fn parse_color_keyword(ident: &str) -> Result<Color, ()> {
     macro_rules! rgb {
@@ -562,26 +602,11 @@ fn parse_color_function<'i, 't, ComponentParser>(
 where
     ComponentParser: ColorComponentParser<'i>,
 {
-    let (red, green, blue, uses_commas) = match_ignore_ascii_case! { name,
+    let (red, green, blue, alpha) = match_ignore_ascii_case! { name,
         "rgb" | "rgba" => parse_rgb_components_rgb(component_parser, arguments)?,
-        "hsl" | "hsla" => parse_hsl_hwb(component_parser, arguments, hsl_to_rgb, /* allow_comma = */ true)?,
-        "hwb" => parse_hsl_hwb(component_parser, arguments, hwb_to_rgb, /* allow_comma = */ false)?,
+        "hsl" | "hsla" => parse_rgb_components_hsl(component_parser, arguments)?,
+        "hwb" => parse_rgb_components_hwb(component_parser, arguments)?,
         _ => return Err(arguments.new_unexpected_token_error(Token::Ident(name.to_owned().into()))),
-    };
-
-    let alpha = if !arguments.is_exhausted() {
-        if uses_commas {
-            arguments.expect_comma()?;
-        } else {
-            arguments.expect_delim('/')?;
-        };
-        clamp_unit_f32(
-            component_parser
-                .parse_number_or_percentage(arguments)?
-                .unit_value(),
-        )
-    } else {
-        255
     };
 
     arguments.expect_exhausted()?;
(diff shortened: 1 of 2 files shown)


emilio:
Ah, sorry, I had missed this one, feel free to explicitly request review so I don't miss it in the future. I'll take a look asap.

src/color.rs
```diff
@@ -262,6 +258,48 @@ pub trait ColorComponentParser<'i> {
...
+
+    /// Parse a `<hue>` value.
+    /// https://w3c.github.io/csswg-drafts/css-color-4/#typedef-hue
+    fn parse_hue<'t>(
+        &self,
+        input: &mut Parser<'i, 't>,
+    ) -> Result<Hue, ParseError<'i, Self::Error>> {
+        let location = input.current_source_location();
+        Ok(match *input.next()? {
+            Token::Number { value: degrees, .. } => Hue { degrees },
```
emilio, line 270: So the reason this (`parse_angle_or_number`, etc) are in `ColorComponentParser` is so that we can add `calc()` support externally e.g., in Firefox. However `Hue` is not constructible outside of this crate so that's not feasible. That needs to at least be fixed

src/color.rs
```diff
@@ -262,6 +258,48 @@ pub trait ColorComponentParser<'i> {
...
+            ref t => return Err(location.new_unexpected_token_error(t.clone())),
+        })
+    }
+
+    /// Parse a `<alpha-value>` value.
+    /// https://w3c.github.io/csswg-drafts/css-color-4/#typedef-alpha-value
+    fn parse_alpha_value<'t>(
+        &self,
+        input: &mut Parser<'i, 't>,
+    ) -> Result<AlphaValue, ParseError<'i, Self::Error>> {
```
emilio, line 293: Same here.

src/color.rs
```diff
@@ -562,109 +602,232 @@ fn parse_color_function<'i, 't, ComponentParser>(
...
-) -> Result<(u8, u8, u8, bool), ParseError<'i, ComponentParser::Error>>
+) -> Result<(u8, u8, u8, u8), ParseError<'i, ComponentParser::Error>>
 where
     ComponentParser: ColorComponentParser<'i>,
 {
-    // Hue given as an angle
-    // https://drafts.csswg.org/css-values/#angles
-    let hue_degrees = component_parser.parse_angle_or_number(arguments)?.degrees();
+    // Try to parse legacy syntax first.
+    let legacy_syntax_result = arguments.try_parse(|input| {
```
emilio, line 742: This refactoring seems somewhat gratuituous? What advantages does it give us? At a glance it seems it would make parsing a bit slower / rewind more.

emilio (changes requested):
Thanks. Not opposed to introducing `Hue` or `AlphaValue` (though I wonder if it'd make sense to do that separately, reusing `parse_number_or_percentage`).

I wonder what's the reasoning for changing the parsing code? At a glance seems it should be slower and without any obvious benefit, but it might have more obvious of benefit with following changes?

In any case we'd need to allow constructing `Hue` / `AlphaValue` externally.

GPHemsley:
> Thanks. Not opposed to introducing `Hue` or `AlphaValue` (though I wonder if it'd make sense to do that separately, reusing `parse_number_or_percentage`).

As I mentioned above, I would have already removed `parse_number_or_percentage()` as redundant if it weren't a public function. The goal of this refactor is to better align concepts and implementation with the css-color-4 spec, and "number or percentage" is not a term defined there. (It's possible that it's a term defined in another spec, but that doesn't seem to have any bearing on `<color>`.)

> I wonder what's the reasoning for changing the parsing code? At a glance seems it should be slower and without any obvious benefit, but it might have more obvious of benefit with following changes?

I laid out my reasoning in https://github.com/servo/rust-cssparser/pull/293#issuecomment-1242908621.

The motivation for this refactoring is to make maintenance easier, based on how the spec is structured. I did not optimize for speed, nor is it clear to me that the difference is significant.

My implementation of `none` will rely on this refactoring more obviously, as `none` is not allowed in legacy syntax and the existing code relies on being able to use the red channel value to determine which branch of the syntax to use.

> In any case we'd need to allow constructing `Hue` / `AlphaValue` externally.

I'm not sure I'm following what you mean here. `parse_hue()` and `parse_alpha_value()` allow you to turn a token stream into their corresponding structs, or you can just specify the values directly.

As far as I can tell, the existing `color` and `calc` implementations in servo would be just as happy to use `Hue` or `AlphaValue` instead, though it's difficult for me to say because servo defines its own `NumberOrPercentage` in multiple places. Can you point me to where you think the implementation needs to be able to differentiate between how the values were derived?

emilio:
My point is, how do we re-implement ColorComponentParser::parse_hue to support e.g. calc() if the hue degrees member is not public and there's no function to construct a Hue? Sorry if I wasn't clear on that point, or if I've missed something.

And yeah, let's remove parse_number_or_percentage. Having a function that gets never called on an interface is silly, and it doesn't particularly help backwards compat (this should be published as a breaking change regardless).

GPHemsley:
> My point is, how do we re-implement ColorComponentParser::parse_hue to support e.g. calc() if the hue degrees member is not public and there's no function to construct a Hue? Sorry if I wasn't clear on that point, or if I've missed something.

I think that's a problem with NumberOrPercentage and AngleOrNumber because they're enums that doesn't exist for Hue and AlphaValue because they're structs.

For example, this is what I've been using to get the values out for tests, which is annoying:
```rust
            if let AngleOrNumber::Angle { degrees } = ColorComponentParser::parse_angle_or_number(
                &DefaultComponentParser,
                &mut Parser::new(&mut ParserInput::new("123deg")),
            )
            .unwrap()
            {
                degrees
            } else {
                unreachable!()
            };
```

But I think for the new stuff we can just do something like this:
```rust
        let x = Hue { degrees: 0. };
        let y = x.degrees;
```

I'll dig around to confirm after I finish with this other thing I'm working on.

> And yeah, let's remove parse_number_or_percentage. Having a function that gets never called on an interface is silly, and it doesn't particularly help backwards compat (this should be published as a breaking change regardless).

Oh, cool. `parse_angle_or_number()`, too, then?

emilio:
> > My point is, how do we re-implement ColorComponentParser::parse_hue to support e.g. calc() if the hue degrees member is not public and there's no function to construct a Hue? Sorry if I wasn't clear on that point, or if I've missed something.
> 
> I think that's a problem with NumberOrPercentage and AngleOrNumber because they're enums that doesn't exist for Hue and AlphaValue because they're structs.
> 
> For example, this is what I've been using to get the values out for tests, which is annoying:
> 
> ```rust
>             if let AngleOrNumber::Angle { degrees } = ColorComponentParser::parse_angle_or_number(
>                 &DefaultComponentParser,
>                 &mut Parser::new(&mut ParserInput::new("123deg")),
>             )
>             .unwrap()
>             {
>                 degrees
>             } else {
>                 unreachable!()
>             };
> ```

Unclear why you wouldn't be able to just use `123` there? That's besides my point anyways. My point is that we want to override `parse_hue` etc, and while you can construct a `NumberOrPercentage` from a dependent crate right now (like [here](https://searchfox.org/mozilla-central/rev/73ff5f1ac45202c4621a1bcee4b35d20027e71ad/servo/components/style/values/specified/calc.rs#788,792)), you wouldn't be able to construct a `Hue` (because visibility of struct fields and enum fields is different). That is, the `degrees` field of `Hue` should be public.

> Oh, cool. `parse_angle_or_number()`, too, then?

Yes, all of them.

GPHemsley:
> > > My point is, how do we re-implement ColorComponentParser::parse_hue to support e.g. calc() if the hue degrees member is not public and there's no function to construct a Hue? Sorry if I wasn't clear on that point, or if I've missed something.
> > 
> > 
> > I think that's a problem with NumberOrPercentage and AngleOrNumber because they're enums that doesn't exist for Hue and AlphaValue because they're structs.
> > For example, this is what I've been using to get the values out for tests, which is annoying:
> > ```rust
> >             if let AngleOrNumber::Angle { degrees } = ColorComponentParser::parse_angle_or_number(
> >                 &DefaultComponentParser,
> >                 &mut Parser::new(&mut ParserInput::new("123deg")),
> >             )
> >             .unwrap()
> >             {
> >                 degrees
> >             } else {
> >                 unreachable!()
> >             };
> > ```
> 
> Unclear why you wouldn't be able to just use `123` there? That's besides my point anyways. 

You mean for the value being parsed? I'm testing angle parsing in this example.

> My point is that we want to override `parse_hue` etc, and while you can construct a `NumberOrPercentage` from a dependent crate right now (like [here](https://searchfox.org/mozilla-central/rev/73ff5f1ac45202c4621a1bcee4b35d20027e71ad/servo/components/style/values/specified/calc.rs#788,792)), you wouldn't be able to construct a `Hue` (because visibility of struct fields and enum fields is different). That is, the `degrees` field of `Hue` should be public.

Oooooh. That was unintentional; I'll fix it. (I think I was actually confusing `struct` with `impl` wrt to visibility.)

> > Oh, cool. `parse_angle_or_number()`, too, then?
> 
> Yes, all of them.

👍🏻

src/color.rs
```diff
@@ -188,13 +181,16 @@ pub enum AngleOrNumber {
...
 
-impl AngleOrNumber {
-    fn degrees(&self) -> f32 {
-        match *self {
-            AngleOrNumber::Number { value } => value,
-            AngleOrNumber::Angle { degrees } => degrees,
-        }
-    }
+/// https://w3c.github.io/csswg-drafts/css-color-4/#hue-syntax
+pub struct Hue {
```
tiaanl, line 185: Perhaps a newtype as there is only one value:

```
pub struct Hue(f32);
```

src/color.rs
```diff
@@ -188,13 +181,16 @@ pub enum AngleOrNumber {
...
+    /// The value as a number of degrees.
+    degrees: f32,
+}
+
+/// https://w3c.github.io/csswg-drafts/css-color-4/#alpha-syntax
+pub struct AlphaValue {
```
tiaanl, line 191: Same as above, also we know it's a value so just `Alpha` would suffice.

src/color.rs
```diff
@@ -262,6 +258,48 @@ pub trait ColorComponentParser<'i> {
             ref t => return Err(location.new_unexpected_token_error(t.clone())),
         })
     }
+
+    /// Parse a `<hue>` value.
+    /// https://w3c.github.io/csswg-drafts/css-color-4/#typedef-hue
+    fn parse_hue<'t>(
```
tiaanl, line 264: This function doesn't really parse hue values, but angles.  Should there be a specific abstraction for hue values only?

src/color.rs
```diff
@@ -562,109 +602,232 @@ fn parse_color_function<'i, 't, ComponentParser>(
...
     ComponentParser: ColorComponentParser<'i>,
 {
-    // Either integers or percentages, but all the same type.
-    // https://drafts.csswg.org/css-color/#rgb-functions
-    let (red, is_number) = match component_parser.parse_number_or_percentage(arguments)? {
-        NumberOrPercentage::Number { value } => (clamp_floor_256_f32(value), true),
-        NumberOrPercentage::Percentage { unit_value } => (clamp_unit_f32(unit_value), false),
-    };
+    // Try to parse legacy syntax first.
+    let legacy_syntax_result = arguments.try_parse(|input| {
```
tiaanl, line 643: This function ends up being very long and duplicates quite a bit of code.  That and the fact that the hsl and hwb versions are also on their own now, duplicating code, I'm not sure this is an improvement over the original.

The abstractions for `Hue` and `Alpha` doesn't add much value on their own.  If you are the consumer of the parse_hue function you don't even know what unit you're getting back.  Adding some unit function might help, e.g. `Hue::as_degrees(..)`, `Hue::as_percentage(..)` or `Hue::as_unit(..)`.  This of course is that `NumberOrPercentage` does with enums.

GPHemsley:
Just to checkpoint this, since it's taking a while...

My initial approach in this PR was to do this refactoring iteratively so as to preserve backwards compatibility.

However, given the feedback above, I have now opted to perform a complete rewrite of color support all at once, based directly on the css-color-4 spec (as opposed to the hybrid between css-color-3 and css-color-4 that currently exists in the codebase). This will allow the addition of a handful of new features straight away and will make it easier to add the remaining support in the future. I am also attempting to draw the line more firmly between parsing and processing, so that more parsing things can be done in this library, but some other processing things are deferred to code that makes use of it.

My work has been complicated by the refactoring that was done as part of #309, but I think I've gotten through that and am now tweaking and tidying up edgecases. That being said, it will likely be a couple more weeks before I have a new PR ready to present. (Until then, I'll leave this one open.)

GPHemsley:
Superseded by #314.

PR #309, opened 2022-10-18
Implement [ok]lab, [ok]lch from color-4 spec.

tiaanl:
https://w3c.github.io/csswg-drafts/css-color-4/#lab-colors

The size of the new color structs are 32*4 bits which is much bigger than the original 32bit for the RGBA value.  This might change in the future.


src/color.rs
```diff
@@ -114,28 +187,275 @@ impl ToCss for RGBA {
...
-            if clamp_unit_f32(rounded_alpha) != self.alpha {
-                rounded_alpha = (self.alpha_f32() * 1000.).round() / 1000.;
+        self.serialize_alpha(dest)?;
+
+        dest.write_char(')')
+    }
+}
+
+// NOTE: LAB and OKLAB is not declared inside the macro, because it causes
+// cbindgen to ignore them.
```
emilio, line 198: Maybe specify which macro? Inside `impl_lab_like` you mean, rigth?

src/color.rs
```diff
@@ -10,6 +10,38 @@ use super::{BasicParseError, ParseError, Parser, ToCss, Token};
 #[cfg(feature = "serde")]
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
 
+/// Implement for colors with an alpha channel.
+pub trait WithAlpha {
```
emilio, line 14: So, `WithAlpha` is mostly an implementation detail for serialization? It seems like we don't really want to expose `serialize_alpha` to consumers, can we either make just `alpha` / `alpha_f32` built-in functions or, if it's useful to have it as a trait (is it?) just keep those, and have a

```
fn serialize_alpha(alpha: u8, dest: &mut impl fmt::Write, legacy_syntax: bool) -> fmt::Result {
```

as a standalone function?

src/color.rs
```diff
@@ -114,28 +187,275 @@ impl ToCss for RGBA {
...
+                W: fmt::Write,
+            {
+                dest.write_str($fname)?;
+                dest.write_str("(")?;
+                self.lightness.to_css(dest)?;
+                dest.write_str(" ")?;
+                self.chroma.to_css(dest)?;
+                dest.write_str(" ")?;
+                self.hue.to_css(dest)?;
+                if self.alpha() != 255 {
```
emilio, line 379: Isn't this check done by `serialize_alpha` anyways?

src/color.rs
```diff
@@ -114,28 +187,275 @@ impl ToCss for RGBA {
...
+        }
+
+        #[cfg(feature = "serde")]
+        impl<'de> Deserialize<'de> for $cls {
+            fn deserialize<D>(deserializer: D) -> Result<Self, serde::de::Error>
+            where
+                D: Deserializer<'de>,
+            {
+                let (lightness, chroma, hue, alpha) = Deserialize::deserialize(deserializer)?;
+                Ok(LCH::new(lightness, chroma, hue, alpha))
```
emilio, line 363: Shouldn't this use `$cls` or `Self::new`?

src/color.rs
```diff
@@ -114,28 +187,275 @@ impl ToCss for RGBA {
...
+        }
+
+        #[cfg(feature = "serde")]
+        impl<'de> Deserialize<'de> for $cls {
+            fn deserialize<D>(deserializer: D) -> Result<Self, serde::de::Error>
+            where
+                D: Deserializer<'de>,
+            {
+                let (lightness, a, b, alpha) = Deserialize::deserialize(deserializer)?;
+                Ok(LAB::new(lightness, a, b, alpha))
```
emilio, line 266: Shouldn't this use `Self::new` rather than hardcoding `LAB`?

src/color.rs
```diff
@@ -114,28 +187,275 @@ impl ToCss for RGBA {
...
+                W: fmt::Write,
+            {
+                dest.write_str($fname)?;
+                dest.write_str("(")?;
+                self.lightness.to_css(dest)?;
+                dest.write_str(" ")?;
+                self.a.to_css(dest)?;
+                dest.write_str(" ")?;
+                self.b.to_css(dest)?;
+                if self.alpha() != 255 {
```
emilio, line 282: Isn't this check done in `serialize_alpha` anyways?

src/color.rs
```diff
@@ -114,28 +187,275 @@ impl ToCss for RGBA {
...
+            fn to_css<W>(&self, dest: &mut W) -> fmt::Result
+            where
+                W: fmt::Write,
+            {
+                dest.write_str($fname)?;
+                dest.write_str("(")?;
+                self.lightness.to_css(dest)?;
+                dest.write_str(" ")?;
+                self.a.to_css(dest)?;
+                dest.write_str(" ")?;
```
emilio, line 280: nit: `write_char(' ')` might be marginally more efficient, bunch of places in this patch.

emilio (changes requested):
The WithAlpha thing is my only concern. I need to take a closer look a the tests but it looks good generally. I think I spotted a couple bugs.

src/color.rs
tiaanl: Simpler with a function for now.  The plan was to replace the `AbsoluteColor::alpha()` function that is used from gecko side.

tiaanl:
Removed the `WithAlpha` for now.  For the tests the only thing I can think is to add some more variations of the values.

src/color.rs
```diff
@@ -106,36 +158,265 @@ impl ToCss for RGBA {
...
+#[repr(C)]
+pub struct LAB {
+    /// The lightness component.
+    pub lightness: f32,
+    /// The a-axis component.
+    pub a: f32,
+    /// The b-axis component.
+    pub b: f32,
+    /// The alpha component.
+    pub alpha: u8,
```
emilio, line 190: I think maybe storing the alpha as an f32 might be better long term (avoids having different computed / animated representations). Wdyt?

src/color.rs
```diff
@@ -106,36 +158,265 @@ impl ToCss for RGBA {
...
+
+impl_lch_like!(LCH, "lch");
+impl_lch_like!(OKLCH, "oklch");
+
+/// An absolutely specified color.
+/// https://w3c.github.io/csswg-drafts/css-color-4/#typedef-absolute-color-base
+#[derive(Clone, Copy, PartialEq, Debug)]
+pub enum AbsoluteColor {
+    /// Specify sRGB colors directly by their red/green/blue/alpha chanels.
+    RGBA(RGBA),
```
emilio, line 357: I think a bit more common would be `Rgba`, `Lab`, `Lch`, `Oklab`, `Oklch` for the enum variants at least, but your call.

emilio (approved):

src/color.rs
tiaanl: That would make things easier yes.  I think I'm going to do it with `RGBA.alpha` as well.  Right now the size of the struct doesn't matter (we'll see if this is an issue) and dealing with alpha values the same would also makes things much simpler.  Can you think of issues with this?

src/color.rs
```diff
@@ -91,51 +144,280 @@ impl Serialize for RGBA {
...
+
+impl_lch_like!(Lch, "lch");
+impl_lch_like!(Oklch, "oklch");
+
+/// An absolutely specified color.
+/// https://w3c.github.io/csswg-drafts/css-color-4/#typedef-absolute-color-base
+#[derive(Clone, Copy, PartialEq, Debug)]
+pub enum AbsoluteColor {
+    /// Specify sRGB colors directly by their red/green/blue/alpha chanels.
+    RGBA(Rgba),
```
emilio, line 358: Actually I meant the variant names should be `Rgba()` etc (so I was expecting `Rgba(RGBA)`, sorry if that wasn't clear). But `Rgba(Rgba)` is also alright.

emilio (approved):
This works for me, with the enum variant names fixed to match the structs.

src/color.rs
emilio: The main issue is that it'd slow down conversion to `nscolor` (the Gecko representation of a computed rgba value which is ~everywhere). But might not be a huge issue.

GPHemsley:
Note that I began a refactor on this same code in #308 that is still awaiting review.

src/color.rs
tiaanl: And now those are renamed too :smile:

src/color.rs
tiaanl: We'll have to convert to rgba (nscolor) anyway from lab/lch/etc. If this becomes a problem we can optimize.

emilio:
(CI fix is in #310. Assuming it's green on top this looks good to me)

PR #311, opened 2022-12-01
Change RGBA components to f32 and change tests to check css.

tiaanl:


tiaanl:
> This is already going to be superseded by the work I'm tracking in #308 (and its successor).

@GPHemsley can you elaborate?  Will your changes override what I've done here?  If so, can you give more details please?

GPHemsley:
@tiaanl I have completely rewritten color.rs to use a data model that more closely aligns with css-color-4, so all of your changes to Rust code are going to be obsolete. (RGBA is going to be relegated to an internal structure used to represent legacy css-color-3 values, which the existing tests use extensively.)

In addition, I am currently wrapping up porting the existing tests to the new data model, so these test changes would disrupt that work. However, I do agree that converting the tests to focus on serialized values is worthwhile, so this isn't entirely for naught.

One sticking point related more to #309 than this, though, is that the tests for CIELab and Oklab colors don't seem to be aligned with (my understanding of) the spec. I'm chasing that down in w3c/csswg-drafts#8065, as the spec leaves a lot to be desired in this area, and WPT currently seems to also be in conflict.

tiaanl:
I replaced this one with #313 which only changes the tests to check against the serialized output.

@GPHemsley is it possible to share your work in progress?  I'm making changes to firefox to implement some color-4 features and would like to avoid double effort going forward.  What I'm working on:

https://bugzilla.mozilla.org/show_bug.cgi?id=1352757
https://bugzilla.mozilla.org/show_bug.cgi?id=1779195
https://bugzilla.mozilla.org/show_bug.cgi?id=1128204

PR #312, opened 2022-12-06
parser: Add parse_comma_separated_ignoring_errors.

emilio:
This would allow to implement
https://bugzilla.mozilla.org/show_bug.cgi?id=1803760 more efficiently.


emilio:
CI failures are https://github.com/rust-lang/rust/issues/105371

jfkthame:
Yeah, having something like this would allow for a more efficient approach -- LGTM.

Might want to add edge-cases like leading, trailing, or successive commas (i.e. completely empty components in the list) to the testcase, just for thoroughness.

PR #313, opened 2022-12-07
Change color tests to check the serialized output

tiaanl:


tiaanl:
@emilio it seems like the optimization you were checking for in #310 is not present on nightly any more.  Should we handle this in a different way?

emilio:
> @emilio it seems like the optimization you were checking for in #310 is not present on nightly any more. Should we handle this in a different way?

Possibly, see https://github.com/rust-lang/rust/issues/105371

emilio:
Maybe we should make ci not block on nightly.

tiaanl:
@emilio It looks like there is no feature that allows marking specific jobs as "allowed to fail". See: https://github.com/actions/toolkit/issues/399

Given what was discussed in the rust-lang thread and not beind able to rely on `Cow` having a specified size, does it make sense to test for the size of this struct?

Also, this is new to me, what are the benefit of having new code compile on beta/nightly and even stable if the msrv is set to a specific version?

emilio:
@tiaanl ok, bumped the msrv and fixed CI again in https://github.com/servo/rust-cssparser/pull/316, can you remove the last couple commits?

PR #314, opened 2022-12-08
WIP: feat(color): Rewrite color in terms of css-color-4

GPHemsley:
(I intend to fully explain this at a later point, but I'm filing this now to facilitate parallel development and the feedback loop. It's still a work in progress.)


src/color.rs
```diff
@@ -150,335 +97,606 @@ impl<'de> Deserialize<'de> for RGBA {
...
-    /// The b-axis component.
-    pub b: f32,
-    /// The alpha component.
-    pub alpha: f32,
+pub struct LabCoords {
+    pub lightness: Option<f32>,
+    pub a: Option<f32>,
+    pub b: Option<f32>,
+    pub alpha: Option<AlphaValue>,
+}
```
emilio, line 149: For the record, I don't think using `Option<f32>` is going to work great for us in terms of representing `none` coordinates. First, it grows the color structs quite considerably.

Second, it's not possible to use it from C/C++, which is a problem for Firefox integration.

Can we do something like this instead?

```
#[repr(transparent)]
pub struct OptionalColorCoordinate(pub f32);

impl OptionalColorCoordinate {
    fn none() -> Self { Self(std::f32::NAN) }
    fn get() -> Option<f32> {
        if self.0.is_nan() { None } else { Some(self.0) }
    }
}
```

Or something along those lines? Same for `Option<AlphaValue>` etc.

src/color.rs
```diff
@@ -150,335 +97,606 @@ impl<'de> Deserialize<'de> for RGBA {
...
+#[derive(Clone, Copy, PartialEq, Debug)]
+pub enum SrgbColor {
+    /// https://w3c.github.io/csswg-drafts/css-color-4/#numeric-srgb
+    Rgb {
+        /// The red channel.
+        red: Option<f64>,
+        /// The green channel.
+        green: Option<f64>,
+        /// The blue channel.
+        blue: Option<f64>,
```
emilio, line 202: Why using f64 here but f32 elsewhere?

emilio (commented):
Some preliminary questions / comments.

cc @tiaanl.

Also, I believe Tiaan was planning on changing the color representation to be a bit more unified, something like:

```rust
struct ColorComponents([f32; 4]);
struct Color(ColorComponents, ColorSpace);
```

Which might have some advantages as you can write interpolation regardless of the color-space etc.

It'd be great if he could weigh in so we could coordinate, to avoid duplicated work.

But yeah, I'd try to avoid growing the `Color` type too much, and it being `#[repr(C)]` or so is kind of a must for us (plus it remaining small is good for performance generally so...).

tiaanl:
> Also, I believe Tiaan was planning on changing the color representation to be a bit more unified, something like:

TLDR: I think the calculation optimized representation (`(color_space, f32, f32, f32, f32)`)should stay in gecko and we just convert to that from whatever the parser gives us.

Right now I take in whatever the parser gives me and transform it into the ~(color_space, f32, f32, f32, f32).  The plan was to move that into the parser once i implement the `color()` function.  I'm not 100% sure if we should do that though.  Not doing so will give us more flexibility to change things for better calculations inside Gecko without changing how things are parsed.  Downside is there are some computation and memory overhead for converting.  Not huge structures, so should be OK, but a benchmark should be done.  As a consumer of this library I would rather get as much detail from the parsed values as some flat optimized representation.

src/color.rs
```diff
@@ -2,36 +2,15 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-use std::f32::consts::PI;
+use std::f64::consts::PI;
```
tiaanl, line 5: I don't think any color representation requires 64-bits of precision.

src/color.rs
```diff
@@ -42,7 +21,7 @@ pub struct RGBA {
     /// The blue component.
     pub blue: u8,
     /// The alpha component.
-    pub alpha: f32,
+    pub alpha: u8,
```
tiaanl, line 24: If `RGBA` is becoming an internal data format is it not better to just remove it completely?

src/color.rs
```diff
@@ -150,335 +97,606 @@ impl<'de> Deserialize<'de> for RGBA {
...
+fn f32_to_f64(f: f32) -> f64 {
+    const PRECISION: f64 = 1e6;
+    (f64::from(f) * PRECISION).round() / PRECISION
+}
 
-        serialize_alpha(dest, self.alpha, true)?;
+/// <hue>
+/// https://w3c.github.io/csswg-drafts/css-color-4/#hue-syntax
+#[derive(Clone, Copy, PartialEq, Debug)]
+pub struct Hue {
```
tiaanl, line 115: maybe ```
type Degrees = f32;
struct Hue(Degrees);
```

Allows for better syntax like `Hue(32.9)`, etc.

File: src/color.rs [hunk 6]
```diff
@@ -150,335 +97,606 @@ impl<'de> Deserialize<'de> for RGBA {
...
-        impl Serialize for $cls {
-            fn serialize<S>(&self, serializer: S) -> Result<serde::ser::Ok, serde::ser::Error>
-            where
-                S: Serializer,
-            {
-                (self.lightness, self.a, self.b, self.alpha).serialize(serializer)
-            }
+/// A color in the sRGB color space.
+#[derive(Clone, Copy, PartialEq, Debug)]
+pub enum SrgbColor {
```
tiaanl, line 194: Will this enum also have the other predefined color spaces? `rec2020`, `linear-srgb`, `display-p3`, etc.?

src/color.rs
GPHemsley: Eventually? Yes, probably. But doing so as part of this PR would have meant a ton more work, so I'd rather leave that for the future.

src/color.rs
GPHemsley: One of the loose ends with this work indeed revolves around the differences in precision between f32 and f64.

On the one hand, 1/255 = 0.0(0392156862745098)..., so it'd be nice to be able to capture the full period of 16 digits.

But more importantly, some of the calculations that are done on these numbers require f64 precision in order to get a clean f32 representation to 6 decimal places (which I would think would be the minimum precision we would aim for).

PI specifically is used to convert degrees to radians.

src/color.rs
GPHemsley: I haven't looked too deeply into the other predefined color spaces to know for sure, but the idea was for each independent color space to have its own enum, with a variant for each way to map into it. (That's why there's `CielabColor` and `OklabColor`.)

So if some of the predefined color spaces map directly into sRGB, then the `SrgbColor` enum would be expanded to include them. But completely independent color spaces like Display P3 and ITU-R BT.2020-2 would get their own enums, and the `Color` enum would get new variants to accommodate them.

But that, of course, is future work.

GPHemsley:
Bah, tests for #312 rely on color parsing.

src/color.rs
GPHemsley: The point about 1/255 is still something to consider, but I otherwise found ways to avoid f64 (including the use of `f32::to_degrees`).

src/color.rs
GPHemsley: @emilio Can you go into more detail about the problems this design causes? I don't like the idea of abusing NaN to capture a concept independent of the float value.

@tiaanl Do you have any thoughts on this?

emilio: I outlined them above:

 * Bigger structs are bad for CSS parsing performance (since they're returned by value quite a lot) and memory usage (if they're stored for long periods of time, which in the case of Firefox they are). Colors are rather common, so it'd be great not to regress parsing performance and memory usage any more than necessary. A struct of four `Option<f32>`s is twice as big as `[f32; 4]`, see [this](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=519921cb0464b5d02b90d97e0b319601), so if there's a not-too-complicated way to halve the size of a color, we should definitely do that.
 * Additionally, we'd love to be able to avoid converting between different color representations between C/C++ and Rust. `Option<>` is not ffi-safe, in general, so we would need to convert colors after parsing to another representation in Firefox if we used it. We _could_ do it if absolutely needed, but if we can avoid it (since it's just boilerplate and potentially duplicated code we'd have to maintain), it'd be great.

Using `NaN` was an idea to deal with both of thos. It allows to encode `none` without wasting extra bits, and it shouldn't make it particularly harder to use from Rust if we do something like what I suggested above, I think? But I'm open to other options.

In an ideal world, there'd be something like a `NotNaN<T: Float>` in the standard library that guarantees the right range so that `Option<NotNaN<f32>>` is guaranteed to be 32 bits, just like `Option<Box<T>>` is guaranteed to be the size of a pointer as long as `T` is `Sized`. That would address both and would be perhaps slightly more idiomatic, but given that doesn't exist, rolling our own as described above seems reasonable?

Another option would be to use something like fixed-point representation where 15 or 31 bits are used for the actual number, and one bit is used for `none`, but that's also a bit more code, and at that point we need to make rather arbitrary precision choices, which would be great to avoid if possible. It's one of the things we could consider doing (using f15 + a bit for `none`) if using f32 for all this becomes prohibitively expensive in terms of memory usage (since it could halve sizes compared to f32 representation). We use that kind of trick in some other places in Firefox like for font-style angles (see [this](https://searchfox.org/mozilla-central/rev/f24503d2f5d8b7e4341be8bc20d94a04a6b223d6/servo/components/style/values/computed/font.rs#845-854)), but again, I'd prefer to avoid it if possible.

Anyways, hopefully that clarifies my concerns with this approach?

GPHemsley: Ah, yes, thank you for that context. I think I understand the issue a bit better now.

I'll try my best to accommodate your request, but one of the things I intended to include in the background explanation that I have yet to write is that I'd like to consider optimizations to be out of scope as much as possible. These changes are unwieldy enough already, and I'd prefer to not add further delay while I fiddle with the idiosyncrasies of Rust; memory management is not my area of expertise. Do we have tests that would be tracking regressions of this nature? And if not, can we commit to adding them in the future?

I'd also like to make explicit one point that I think is already implicit in your comment: We want to be sure to differentiate between optimizations (and other things) that are broadly applicable and those that are Firefox-specific. The former are definitely appropriate for this repo, but the latter probably aren't.

src/color.rs
GPHemsley: I intentionally implemented it the way I did to serve two purposes:
* Avoid having to use `.0` to get at the value.
* Make clear that the value has been normalized from what may have been originally specified (e.g. radians).

So needing to use `Hue::new(32.9)` was an intentional design decision.

I don't know if that changes your opinion?

src/color.rs
emilio: > Do we have tests that would be tracking regressions of this nature?

Sure, we do have CSS parsing benchmarks for Firefox like [this](https://searchfox.org/mozilla-central/source/layout/style/test/gtest/StyloParsingBench.cpp), plus other performance tests that are monitored regularly.

That said, those obviously benchmark more than color parsing, and need a Firefox checkout plus incorporating your changes into Firefox which I don't think is something reasonable to ask of you.

But it should be rather trivial to add a few benchmarks for color parsing in `src/tests.rs` like we have for `data:` uris etc?

> We want to be sure to differentiate between optimizations (and other things) that are broadly applicable and those that are Firefox-specific. The former are definitely appropriate for this repo, but the latter probably aren't.

Sure, I agree. Firefox has the extra requirements of dealing with C++ and thus needing ffi-safe types etc. I'd be ok merging a change that requires Firefox to support ffi on top of cssparser (even though I think it'd be unfortunate).

However general parsing performance should be applicable to all other rust crates that depend on `cssparser`. Firefox might have stricter performance requirements than some other programs, but I don't think I'd be fine merging a change with performance characteristics that don't work for Firefox.

GPHemsley:
A number of spec changes have been made as a result of this work, but it does likely mean portions of the code will need to be changed (hopefully to simplify them).

PR #317, opened 2022-12-13
Add parsing of color()

tiaanl:
Add parsing of predefined color spaces with the 'color(<color_space> ...)' function.

Tests are based on 'css/css-color/parsing/color-valid-color-function.html' from the web platform tests.


src/color.rs
```diff
@@ -350,6 +350,168 @@ macro_rules! impl_lch_like {
...
+            PredefinedColorSpace::XyzD65 => "xyz-d65",
+        }
+    }
+}
+
+impl std::str::FromStr for PredefinedColorSpace {
+    type Err = ();
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        Ok(match s.to_lowercase().as_str() {
```
emilio, line 395: let's use `match_ignore_ascii_case` instead.

src/color.rs
```diff
@@ -350,6 +350,168 @@ macro_rules! impl_lch_like {
...
+    fn to_css<W>(&self, dest: &mut W) -> fmt::Result
+    where
+        W: fmt::Write,
+    {
+        dest.write_str(self.as_str())
+    }
+}
+
+#[cfg(feature = "serde")]
+impl<'de> Deserialize<'de> for PredefinedColorSpace {
```
emilio, line 420: I don't understand, can't this be derived? Serializing as a string looks rather inefficient

src/color.rs
```diff
@@ -350,6 +350,168 @@ macro_rules! impl_lch_like {
...
+            color_space,
+            c1,
+            c2,
+            c3,
+            alpha,
+        }
+    }
+}
+
+#[cfg(feature = "serde")]
```
emilio, line 474: Same, this seems like it could be derived.

src/color.rs
```diff
@@ -350,6 +350,168 @@ macro_rules! impl_lch_like {
...
+}
+
+impl ToCss for ColorFunction {
+    fn to_css<W>(&self, dest: &mut W) -> fmt::Result
+    where
+        W: fmt::Write,
+    {
+        dest.write_str("color(")?;
+        self.color_space.to_css(dest)?;
+        dest.write_str(" ")?;
```
emilio, line 502: `write_char(' ')`

src/color.rs
```diff
@@ -1072,3 +1240,55 @@ where
...
+        Ok(match_ignore_ascii_case! {
+            ident,
+            "srgb" => PredefinedColorSpace::Srgb,
+            "srgb-linear" => PredefinedColorSpace::SrgbLinear,
+            "display-p3" => PredefinedColorSpace::DisplayP3,
+            "a98-rgb" => PredefinedColorSpace::A98Rgb,
+            "prophoto-rgb" => PredefinedColorSpace::ProphotoRgb,
+            "rec2020" => PredefinedColorSpace::Rec2020,
+            "xyz-d50" => PredefinedColorSpace::XyzD50,
+            "xyz" | "xyz-d65" => PredefinedColorSpace::XyzD65,
```
emilio, line 1264: Let's use `from_str`, or remove `from_str` once it's not used for serialization?

emilio (changes requested):

src/color.rs
tiaanl: If I derive on `PredefinedColorSpace` then I have to add `#[cfg_attr(feature = serde, serde(rename = "???"))]` to every item in the enum.  I still want the `FromStr`, but I don't want to `serde::deserialize` there, i just want a match.  I don't want to maintain 2 lists of constants.

If you say serialize to string is inefficient, do you mean serialize to something else?

tiaanl: Also, I'm not even sure we need serialization for the color struct any more as we're matching against css for testing now. But another issue I guess.

emilio: I'm confused, why do we need the serialization anymore? It was useful for tests iirc.

Now we don't need it, and we don't allow serializing the whole `Color` struct anyways. So I'd rather get rid of it.

Specially since serializing from / to string is rather inefficient (color space should just be serialized as an integer for the discriminant).

src/color.rs
```diff
@@ -1072,3 +1245,45 @@ where
...
+) -> Result<Color, ParseError<'i, ComponentParser::Error>>
+where
+    ComponentParser: ColorComponentParser<'i>,
+{
+    let location = arguments.current_source_location();
+    let color_space = arguments.try_parse(|i| {
+        let ident = i.expect_ident()?;
+        PredefinedColorSpace::from_str(ident)
+            .map_err(|_| location.new_unexpected_token_error(Token::Ident(ident.clone())))
+    })?;
```
emilio, line 1262: I don't think you need the `try_parse` here, if you don't parse a correct color space you just return an error either way, so let's remove it.

src/color.rs
```diff
@@ -1072,3 +1245,45 @@ where
...
+
+    macro_rules! parse_component {
+        () => {{
+            if let Ok(c) = arguments.try_parse(|i| component_parser.parse_number_or_percentage(i)) {
```
emilio, line 1266: Why is this optional? This is allowing to parse `color(srgb)` which seems wrong per spec. We eventually need to support `none`, but that's a different issue.

emilio: I believe this should just do:

```
let c1 = component_parser.parse_number_or_percentage(arguments)?;
...
```

unless I'm missing something?

emilio (changes requested):

src/color.rs
tiaanl: Nice catch, removing the `try_parse`

src/color.rs
tiaanl: Removed serialization where we don't use it in tests.

emilio (approved):

PR #325, opened 2023-02-17
Parser takes "none" keyword into account

tiaanl:
"none" values will not be parsed where allowed and Option<T> values will be passed through to the FromParsedColor trait to construct new colors.

The default color type (Color) also now allows Option<T> values in it's members.


tiaanl:
I split this MR into parts, so closing this one.

PR #326, opened 2023-02-26
Do not convert hsl/hwb to RGB at parse time to not loose precision

tiaanl:
Parsing is not done in 2 forms as per the spec.  Legacy and new syntax.  For rgb and hsl where both are valid, first one is tried and if it fails, we try the other.

We also store hsl and hwb values directly without converting them to rgba at parse time.  The conversion only happens how before serializing.

### DIFF
--- a/src/color.rs
+++ b/src/color.rs
@@ -182,6 +182,122 @@ impl ToCss for RGBA {
     }
 }
 
+#[derive(Clone, Copy, PartialEq, Debug)]
+pub struct Hsl {
+    /// The hue component.
+    pub hue: f32,
+    /// The saturation component.
+    pub saturation: f32,
+    /// The lightness component.
+    pub lightness: f32,
+    /// The alpha component.
+    pub alpha: f32,
+}
+
+impl Hsl {
+    pub fn new(hue: f32, saturation: f32, lightness: f32, alpha: f32) -> Self {
+        Self {
+            hue,
+            saturation,
+            lightness,
+            alpha,
+        }
+    }
+}
+
+impl ToCss for Hsl {
+    fn to_css<W>(&self, dest: &mut W) -> fmt::Result
+    where
+        W: fmt::Write,
+    {
+        // HSL serializes to RGB, so we have to convert it.
+        let (red, green, blue) = hsl_to_rgb(
+            self.hue / 360.0, // Hue is expected in range [0..1].
+            self.saturation,
+            self.lightness,
+        );
+
+        RGBA::from_floats(red, green, blue, self.alpha).to_css(dest)
+    }
+}
+
+#[cfg(feature = "serde")]
+impl Serialize for Hsl {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        (self.hue, self.saturation, self.lightness, self.alpha).serialize(serializer)
+    }
+}
+
+#[cfg(feature = "serde")]
+impl<'de> Deserialize<'de> for Hsl {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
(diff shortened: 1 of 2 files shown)


src/color.rs
```diff
@@ -1060,87 +1149,149 @@ fn parse_rgb<'i, 't, P>(
...
-        }
-        blue = clamp_floor_256_f32(color_parser.parse_number(arguments)?);
-    } else {
-        green = clamp_unit_f32(color_parser.parse_percentage(arguments)?);
-        if uses_commas {
-            arguments.expect_comma()?;
+    // 0 = none
+    // 1 = number
+    // 2 = percentage
+    let mut types = vec![];
```
emilio, line 1183: I'm pretty sure having an extra heap-allocation here is not acceptable... Why can't we keep the rgb code as it was? It's the hottest one by far, and it doesn't seem it gives us much factoring out if we then need to filter the types manually. Can we split the component parsing changes from the `Hsl` / `Hwb` changes?

src/color.rs
```diff
@@ -1060,87 +1149,149 @@ fn parse_rgb<'i, 't, P>(
...
-    // https://drafts.csswg.org/css-color/#rgb-functions
-    let (red, is_number) = match color_parser.parse_number_or_percentage(arguments)? {
-        NumberOrPercentage::Number { value } => (clamp_floor_256_f32(value), true),
-        NumberOrPercentage::Percentage { unit_value } => (clamp_unit_f32(unit_value), false),
+    let location = arguments.current_source_location();
+
+    // rgb(..) and rgba(..) allows both legacy and new syntax, so try legacy
+    // first and if that fails, try the newer syntax. The newer syntax still
+    // does not support "none".
+    let start = arguments.state();
```
emilio, line 1157: `input.try_parse` basically does this, right?

src/color.rs
```diff
@@ -1060,87 +1149,149 @@ fn parse_rgb<'i, 't, P>(
...
+        NumberOrPercentage::Percentage { .. } => types.push(2),
+    }
+    match blue {
+        NumberOrPercentage::Number { .. } => types.push(1),
+        NumberOrPercentage::Percentage { .. } => types.push(2),
+    }
+    if types.len() > 1 {
+        let first = *types.first().unwrap(); // Already checked that the vec is not empty.
+        if !types.into_iter().all(|i| i == first) {
+            return Err(location.new_unexpected_token_error(Token::Ident("".into())));
```
emilio, line 1199: Hm? This seems like a hack...

src/color.rs
```diff
@@ -1289,3 +1423,69 @@ where
...
+    input: &mut Parser<'i, 't>,
+    f1: F1,
+    f2: F2,
+    f3: F3,
+) -> Result<(R1, R2, R3, f32), ParseError<'i, P::Error>>
+where
+    P: ColorParser<'i>,
+    F1: Fn(&P, &mut Parser<'i, 't>) -> Result<R1, ParseError<'i, P::Error>>,
+    F2: Fn(&P, &mut Parser<'i, 't>) -> Result<R2, ParseError<'i, P::Error>>,
+    F3: Fn(&P, &mut Parser<'i, 't>) -> Result<R3, ParseError<'i, P::Error>>,
```
emilio, line 1441: These should probably be `FnOnce`

emilio (changes requested):
So would there be any chance of getting more context on why the color component parsing changes are an improvement? At a glance, it makes parsing potentially slower for non-legacy colors, and the extra heap allocation during rgb color parsing seems unfortunate, we should be able to do better there.

Can we split that from the hsl/hwb changes? I assume the split you're doing makes sense when we start parsing `none`, but as is I fail to see why it's an improvement.

src/color.rs
tiaanl: Parsing changes are now in a separate PR.

emilio (approved):
Looks good, let's fix the PR title?

PR #327, opened 2023-02-26
Add none keyword

tiaanl:
For each color component, we allow the "none" keyword.  The keyword will be kept in place for serialization.  The "none" keyword is not allowed in the legacy syntax.

`rgb` and `hsl` are handled special, because they support the legacy color syntax.  They also cannot be combined, because rgb cares about the type of the components, and hsl does not.

### DIFF
--- a/src/color.rs
+++ b/src/color.rs
@@ -1269,6 +1269,21 @@ where
     })
 }
 
+fn parse_modern_alpha<'i, 't, P>(
+    color_parser: &P,
+    arguments: &mut Parser<'i, 't>,
+) -> Result<Option<f32>, ParseError<'i, P::Error>>
+where
+    P: ColorParser<'i>,
+{
+    if !arguments.is_exhausted() {
+        arguments.expect_delim('/')?;
+        parse_none_or(arguments, |p| parse_alpha_component(color_parser, p))
+    } else {
+        Ok(Some(OPAQUE))
+    }
+}
+
 #[inline]
 fn parse_rgb<'i, 't, P>(
     color_parser: &P,
@@ -1324,12 +1339,7 @@ where
             color_parser.parse_number_or_percentage(p)
         })?);
 
-        if !arguments.is_exhausted() {
-            arguments.expect_delim('/')?;
-            parse_none_or(arguments, |p| parse_alpha_component(color_parser, p))?.unwrap_or(0.0)
-        } else {
-            OPAQUE
-        }
+        parse_modern_alpha(color_parser, arguments)?.unwrap_or(0.0)
     };
 
     Ok(P::Output::from_rgba(red, green, blue, alpha))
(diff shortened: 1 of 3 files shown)


src/color.rs
```diff
@@ -1187,32 +1277,60 @@ fn parse_rgb<'i, 't, P>(
...
+            color_parser.parse_number_or_percentage(p)
+        })?);
+
+        blue = get_component_value(&parse_none_or(arguments, |p| {
+            color_parser.parse_number_or_percentage(p)
+        })?);
+
+        if !arguments.is_exhausted() {
+            arguments.expect_delim('/')?;
+            parse_none_or(arguments, |p| parse_alpha_component(color_parser, p))?.unwrap_or(0.0)
```
emilio, line 1329: I'm confused, why don't we need to preserve `None` in `rgba`? Seems to me we should (even if it should serialize when resolved as zero).

tiaanl:
Looks like this is causing quite a few errors as is, need to investigate.

src/color.rs
tiaanl: Its useless information right now.  Perhaps when relative color syntax is added it is needed, but I haven't looked so far into that.  Any particular reason you think it should be added even if its not used?

emilio: You need it for interpolation don't you?

`color-mix(in srgb, rgba(100 100 100 / 1.0) 50%, rgba(0 none 0 / 1.0))` should be `rgba(50, 100, 50, 1.0)` not `rgba(50, 50, 50, 1.0)`, right?

src/color.rs
```diff
@@ -63,101 +63,49 @@ fn normalize_hue(hue: f32) -> f32 {
...
+        green: Option<f32>,
+        blue: Option<f32>,
+        alpha: Option<f32>,
+    ) -> Self {
         Self::new(
-            clamp_unit_f32(red),
-            clamp_unit_f32(green),
-            clamp_unit_f32(blue),
-            alpha.max(0.0).min(1.0),
+            red.map(|r| clamp_unit_f32(r)),
```
emilio, line 87: nit: Just `.map(clamp_unit_f32)`, here and elsewhere.

src/color.rs
```diff
@@ -892,6 +840,42 @@ pub trait FromParsedColor {
     ) -> Self;
 }
 
+/// Parse a color hash, without the leading '#' character.
+#[inline]
+
```
emilio: nit: Stray newline

emilio (approved):
I only looked at the new changes but seems reasonable. r=me with the bits below addressed.

emilio:
@bors-servo delegate+

(@tiaanl this allows you to `at-bors-servo r=emilio` when comments are addressed)

src/color.rs
tiaanl: Added `none` support to rgb and hsl as well.

PR #328, opened 2023-03-09
Combine parsing into legacy and new parsing.

tiaanl:
Combine the rules of legacy and new parsing into single functions.  This allows for a much clearer separation of which keywords support which syntax.  RGB is still special, because it the components **must** be of the same type (numbers or percentages).  Most of the color functions are using the new syntax, but some special notes:

- `RGB` uses it's own parser as mentioned.
- `HSL` supports both legacy and new syntax.
- `HWB` only supports new suntax.
- _ the rest supports the new syntax only.

These changes are also in preparation of introducing the `none` keyword where it is much more important to know which syntax is being used, as legacy syntax does **not** support the `none` keyword.

### DIFF
--- a/src/color.rs
+++ b/src/color.rs
@@ -540,10 +540,10 @@ impl NumberOrPercentage {
         }
     }
 
-    fn value(&self, percentage_max: f32) -> f32 {
+    fn value(&self, percentage_basis: f32) -> f32 {
         match *self {
             Self::Number { value } => value,
-            Self::Percentage { unit_value } => unit_value * percentage_max,
+            Self::Percentage { unit_value } => unit_value * percentage_basis,
         }
     }
 }
--- a/src/color.rs
+++ b/src/color.rs
@@ -32,6 +32,13 @@ fn serialize_alpha(dest: &mut impl fmt::Write, alpha: f32, legacy_syntax: bool) 
     rounded_alpha.to_css(dest)
 }
 
+// Guaratees hue in [0..360)
+fn normalize_hue(hue: f32) -> f32 {
+    // https://drafts.csswg.org/css-values/#angles
+    // Subtract an integer before rounding, to avoid some rounding errors:
+    hue - 360.0 * (hue / 360.0).floor()
+}
+
 /// A color with red, green, blue, and alpha components, in a byte each.
 #[derive(Clone, Copy, PartialEq, Debug)]
 #[repr(C)]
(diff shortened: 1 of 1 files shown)


src/color.rs
```diff
@@ -1289,3 +1287,61 @@ where
...
+    P: ColorParser<'i>,
+    F1: FnOnce(&P, &mut Parser<'i, 't>) -> Result<R1, ParseError<'i, P::Error>>,
+    F2: FnOnce(&P, &mut Parser<'i, 't>) -> Result<R2, ParseError<'i, P::Error>>,
+    F3: FnOnce(&P, &mut Parser<'i, 't>) -> Result<R3, ParseError<'i, P::Error>>,
+{
+    let r1 = f1(color_parser, input)?;
+    let r2 = f2(color_parser, input)?;
+    let r3 = f3(color_parser, input)?;
+
+    let alpha = if !input.is_exhausted() {
```
emilio, line 1339: Seems we could keep having a single `parse_alpha`, but pass `/* uses_commas = */ false` here?

src/color.rs
```diff
@@ -1289,3 +1287,61 @@ where
         alpha,
     ))
 }
+
+/// Try to parse the components and alpha with the legacy syntax, but also allow
+/// the [color-4] syntax if that fails.
+/// https://drafts.csswg.org/css-color-4/#color-syntax-legacy
+pub fn parse_legacy_components<'i, 't, P, F1, F2, F3, R1, R2, R3>(
```
emilio, line 1294: Maybe `parse_components_allow_legacy` would be a better name for this?

(But then again we could have a single `parse_components` with `allow_legacy: bool`). I assume this makes it easier to handle `none` in the future somehow tho, so this is fine as an interim step.

src/color.rs
```diff
@@ -532,6 +539,13 @@ impl NumberOrPercentage {
             NumberOrPercentage::Percentage { unit_value } => unit_value,
         }
     }
+
+    fn value(&self, percentage_max: f32) -> f32 {
```
emilio, line 543: I think the common name for this would be `percentage_basis`. But this is fine too.

emilio (approved):
I like the `parse_components` split. If this PR is on its own was the final state, I'd insist on having a simple `parse_components` with an `allow_legacy` argument, same for `parse_alpha` with `uses_commas`.

But if you somehow require the split later down the road and this makes it easier for you, then this is fine with the nits below.

src/color.rs
tiaanl: Cool, wasn't sure what to name it :smile:

src/color.rs
tiaanl: Yes, adding `none` complicates things, because the legacy syntax doesn't allow `none`.  But also, because if the function supports legacy, then the newer syntax for that same function doesn't allow `none` either.  So adding the `allow_legacy` turns into an if explosion.

emilio:
@bors-servo r+

Okay, let's try this. if we care about the extra branches when we know we can't parse legacy syntax, I think a template argument `const ALLOW_LEGACY: bool` instead of two functions should make sure the compiler can optimize those out (if it can't already).

But anyways let's do this for now and we can reevaluate as we implement `none`.

PR #329, opened 2023-03-11
Cleanup for crates.io release

tiaanl:
- Add documentation where needed.
- Remove #[repr(C)] from structs.  These are not needed any more and doesn't really do anything now that the types are Option<_>.
- Fix links so that they work in docs.
- Small clippy fixes.

@emilio For publishing all the new color changes as 0.30 to crates.io, any other suggestions on what to clean up?


emilio (approved):
Sorry I missed this one. Looks good

PR #331, opened 2023-04-27
Minor rules and declaration cleanups.

emilio:


emilio:
@tiaanl @SimonSapin @jdm any objection on bumping MSRV to 1.63 which fixes the bugs that we were working around in #254?

Seems already old-ish enough, and allows #256 to get through too. I'm going to rejigger some of this code for nesting and it'd be good not having to think about the weird lifetime inference bug.

SimonSapin:
No objection!

emilio:
@bors-servo r+

Alright then :)

emilio:
Well actually, since there's a release pending, let me do that first. @bors-servo r-

PR #333, opened 2023-04-27
[css-nesting] Trivial-ish nesting API changes.

emilio:
The API change is needed regardless (we need a declaration parser that also parses qualified rules).

The implementation details are a bit subject to change so make it opt-in.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "cssparser"
-version = "0.30.0"
+version = "0.31.0"
 authors = [ "Simon Sapin <simon.sapin@exyr.org>" ]
 
 description = "Rust implementation of CSS Syntax Level 3"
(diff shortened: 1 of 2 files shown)


emilio:
r? @tiaanl @SimonSapin 

I'm pretty sure we need the API change (requiring `DeclarationListParser` to implement `QualifiedRuleParser`), so I want to land it now (unpublished) to keep working on the Gecko bits.

The main question tho, is whether we want to unify these APIs a bit, which might be possible (something effectively like a trait `RuleAndDeclarationParser` which perhaps switches behavior a bit based on stylesheet position). But maybe for the future, once we have a better idea of what the nesting syntax requirements end up being?

SimonSapin:
This crate hasn’t exactly been the most stable in terms of avoiding breaking changes to public APIs, so I think another one is fine.

I’m not planning to do more maintenance work (beyond my recent PR) and I don’t have a strong opinion for this API, so I’ll leave it to y’all to decide.

tiaanl (approved):

PR #335, opened 2023-04-30
Encapsulate unsafe in `_cssparser_internal_to_lowercase` macro

nicoburns:
# Objective

- Allow the `_cssparser_internal_to_lowercase` macro to be used in crates that set `#[forbid(unsafe)]`.
- Document the safety invariant for this unsafe block

## Changes made

- Wrap the `unsafe` block in the `_cssparser_internal_to_lowercase` macro in an `#[inline(always)]` function.
-  Add comments

## Context

- https://github.com/DioxusLabs/taffy/pull/460#discussion_r1174152707


SimonSapin:
Oh, good idea! This is now possible with const generics

@bors-servo r+

Issue #128, opened 2017-03-13
Remove match_byte macro when Rust 1.17 is stable.


SimonSapin:
@nox pointed out (https://github.com/servo/rust-cssparser/pull/112#issuecomment-286152329) that https://github.com/rust-lang/rust/pull/39456 brings the same performance to plain `match` expressions.

SimonSapin:
1.17.0 is stable, but the `tests::unquoted_url` benchmark is ~2x slower when compiled with `--features dummy_match_byte`. CC @nox, @emilio

nox:
Sad!

SimonSapin:
Still a ~2x difference in 1.71

PR #337, opened 2023-05-08
build: Make match_byte a proper proc-macro.

emilio:
Now these are stable. Fix a bunch of warnings that this exposed.

This seems still a valuable optimization from looking at existing benchmarks.


src/tokenizer.rs
```diff
@@ -9,6 +9,7 @@ use crate::cow_rc_str::CowRcStr;
 use crate::parser::ParserState;
 use std::char;
 use std::ops::Range;
+use cssparser_macros::match_byte;
```
SimonSapin, line 12: Instead of adding support for `dummy_match_byte` in the proc macro I think it’d be nicer to keep the cfg-conditional macro_rule and make this import conditional with the opposite cfg

SimonSapin (commented):
Making `cargo fmt` and rust-analyzer Just Work will be a nice improvement, and yes the optimization is still relevant: https://github.com/servo/rust-cssparser/issues/128#issuecomment-1539038376

src/tokenizer.rs
emilio: Sure, sounds good, done.

SimonSapin (approved):

PR #338, opened 2023-05-14
[css-nesting] Remove DeclarationListParser.

emilio:
For nesting, we need to parse declaration and nested rules together. Rather than having DeclarationListParser, organize the parsing code in two:

 * StyleSheetParser for the top level stylesheet, which has the relevant hacks and so on.
 * RuleBodyParser which does potentially everything else.

For that, introduce a new trait that agglomerates all the nesting needs (so potentially all of at-rule + qualified-rule + declaration parsing).

While at it, change the API a little bit so that these take mutable
references, so that we can avoid some copying.


tiaanl (approved):
Haven't looked at the implementation of this in fx, but no issues here.

PR #339, opened 2023-05-16
[css-nesting] Fix body parsing recovery when nesting is not enabled.

emilio:
Without the new nesting proposal, a stray semicolon in a media rule eats the next rule.


emilio:
I'll experiment a bit more with such a setup once I get nesting working... Maybe it's simpler.

PR #340, opened 2023-05-19
Make the serialize_alpha function public.

tiaanl:
This function is very handy in other places and is worth making public.  Right now it will be used to generate specified value colors in Gecko.

### DIFF
--- a/src/color.rs
+++ b/src/color.rs
@@ -26,9 +26,10 @@ where
     }
 }
 
+/// Serialize the alpha copmonent of a color according to the specification.
 /// <https://drafts.csswg.org/css-color-4/#serializing-alpha-values>
 #[inline]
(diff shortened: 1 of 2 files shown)


src/color.rs
```diff
@@ -26,9 +26,10 @@ where
     }
 }
 
+/// Serialize the alpha copmonent of a color according to the specification.
 /// <https://drafts.csswg.org/css-color-4/#serializing-alpha-values>
 #[inline]
-fn serialize_alpha(
+pub fn serialize_alpha(
```
emilio, line 32: Let's call this serialize_color_alpha or something then, since there's a other alpha values like opacity that are just a plain number / percent.

emilio (approved):
Looks good with that rename or similar.

PR #343, opened 2023-06-02
[css-nesting] Make our nesting implementation match the new spec better.

emilio:
This fixes issues that were found when enabling nesting by default on Nightly.

It matches the spec (see new link) more closely. The places where we diverge (continuing rather than stopping when hitting right-curly, etc) are not really observable because of how our other APIs work (parse_nested_block in particular, for that case).


tiaanl (approved):
Looks good.

PR #350, opened 2023-07-20
Run miri on CI

### LINKED ISSUE #346: Set up miri on CI
Once https://github.com/upsuper/dtoa-short/pull/8 and https://github.com/servo/rust-cssparser/pull/345 are fixed miri is green. We should probably skip some slow tests under miri but it should be feasible to set up miri on CI other than that.

zrhoffman:
Resolves #346. Some tests that take over a minute to run under miri are skipped (on CI only):

* color3_hsl
* color4_hwb
* color4_lab_lch_oklab_oklch
* no_stack_overflow_multiple_nested_blocks
* numeric
* roundtrip_percentage_token

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -53,7 +53,7 @@ jobs:
 
       - name: Cargo miri test
         if: "matrix.toolchain == 'nightly'"
-        run: cargo miri test ${{ matrix.features }}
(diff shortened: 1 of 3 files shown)


zrhoffman:
- Rebased onto latest master
- Using a feature to skip the slower tests under miri on CI

emilio (approved):

PR #351, opened 2023-07-20
Enhance modern component serialization with inf and nan support

tiaanl:
Modern syntax colors support NaN and infinity values. These values are now allowed, where previously for some lab/lch/oklab/oklch components, NaN values were truncated to 0.0.

NaN and infinity now also correctly serializes to calc(NaN) and calc(infinity) respectively.


emilio (commented):
This seems related to https://github.com/web-platform-tests/interop/issues/369 / https://github.com/w3c/csswg-drafts/issues/8629 which is still under discussion. Could you elaborate on the behavior you're implementing?

In particular, the clamping infinity but not `NaN` at specified value time seems like a bug?

tiaanl:
Ok now without any of the adjusting or clamping.

src/color.rs
```diff
@@ -55,6 +45,33 @@ pub fn serialize_color_alpha(
...
+
+impl<'a> ToCss for ModernComponent<'a> {
+    fn to_css<W>(&self, dest: &mut W) -> fmt::Result
+    where
+        W: fmt::Write,
+    {
+        if let Some(value) = self.0 {
+            if value.is_infinite() {
+                if value.is_sign_negative() {
+                    dest.write_str("calc(-infinity)")
```
emilio, line 60: Hmm, so the behavior of out-of-range negative values like `-3` would be different than `-infinity`? Or would it be consistent?

src/color.rs
```diff
@@ -1591,3 +1608,45 @@ where
 
     Ok((r1, r2, r3, alpha))
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn serialize_modern_components() {
```
emilio, line 1617: (Not convinced this test is particularly useful but...)

emilio:
Ah I guess we're not doing the bors thing. Anyways please confirm this is consistent and hit merge yourself? I think you can, if not happy to do that :)

src/color.rs
tiaanl: `-3` would serialize to `-3` and `-infinity` would serialize to `calc(-infinity)`.

As far as the parser is concerned now, there is no range, just numbers between -inf..inf.  Up to the user of the lib to clamp/decide what is a range.

src/color.rs
tiaanl: What would you change to be convinced?

src/color.rs
emilio: Ok, we should make sure that Gecko is also consistent in how we treat these but that seems fine.

PR #352, opened 2023-07-24
Rgba does not store Option as component

tiaanl:
Legacy `rgb`/`rgba` syntax does not allow the "none" keyword and although the modern `rgb`/`rgba` syntax does support it, it is immediately converted to 0, because we never have to serialize back to "none" keywords for `rgb`/`rgba`.

Also `RGBA` is now renamed to `Rgba` as per common rust naming conventions now that it is not used directly in Gecko any more.


src/color.rs
```diff
@@ -82,44 +82,34 @@ fn normalize_hue(hue: f32) -> f32 {
 
 /// A color with red, green, blue, and alpha components, in a byte each.
 #[derive(Clone, Copy, PartialEq, Debug)]
-pub struct RGBA {
+pub struct Rgba {
```
emilio, line 85: Maybe we should call this `LegacyRgba` or somesuch?

emilio (approved):

src/color.rs
tiaanl: :+1: Makes more sense, fixing.

PR #353, opened 2023-08-02
color: Expose named colors more directly.

emilio:
This will allow me to remove some duplicated version of this from Gecko.

This changes the API of ascii_case_insensitive_phf_map, so needs to be a breaking bump.


tiaanl (approved):
Looks good.

PR #356, opened 2023-08-24
tokenizer: Source positions are always at char boundaries, so avoid overhead when slicing

emilio:
These show up in profiles: https://share.firefox.dev/45Gh2s7


tiaanl (approved):
Hope the debug_asserts doesn't make debug builds too slow.

emilio:
Nah, they were checked in the index operation before.

PR #357, opened 2023-08-31
Make NumberOrPercentage and AngleOrNumber getters public.

tiaanl:
Because the structs are public it only make sense to make the getters public as well.

### DIFF
--- a/src/color.rs
+++ b/src/color.rs
@@ -669,14 +669,17 @@ pub enum NumberOrPercentage {
 }
 
 impl NumberOrPercentage {
-    fn unit_value(&self) -> f32 {
+    /// Return the value as a percentage.
+    pub fn unit_value(&self) -> f32 {
         match *self {
             NumberOrPercentage::Number { value } => value,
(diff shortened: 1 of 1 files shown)


src/color.rs
```diff
@@ -669,14 +669,17 @@ pub enum NumberOrPercentage {
 }
 
 impl NumberOrPercentage {
-    fn unit_value(&self) -> f32 {
+    /// Return the value as a percentage.
+    pub fn unit_value(&self) -> f32 {
```
emilio, line 673: I think in the calc() code we call this `unitless_value`, maybe do the same for consistency?

emilio (approved):

src/color.rs
tiaanl: `unit_value` comes from `Token::Percentage { unit_value, .. }`. Consistency can go either way here, so without having to swap the two methods around, because unitless value will now be the "other" one, I'd rather keep it as is for now.

PR #360, opened 2023-09-06
parser: minor clean-ups.

emilio:


emilio:
(This is rather trivial, so will merge since it blocks #359)

PR #361, opened 2023-09-06
serializer: Use match_byte in the serializer.

emilio:


tiaanl (approved):
+256 bytes of data per call site is unfortunate as the data is very sparse and mostly 0.

I wonder how the performance would be affected if you partition the lookup table.  Say 16 bytes each, not creating blocks that are all 0, and then have a single u16 that has bits set for each "existing" block.  If you hit a "non existent" block, you default out otherwise do the table lookup.  No for this PR though :smile:

PR #369, opened 2023-11-09
Replace TokenSerializationType with TokenSerializationTypeVariants

### LINKED ISSUE #266: No way to get TokenSerializationType directly
From https://bugzilla.mozilla.org/show_bug.cgi?id=1503656

Actually there is no way to get `TokenSerializationType` directly from `Token` type. If it is `TokenSerializationTypeVariants::Nothing`, it has `nothing()` function to get it.

Even parsing CSS that is hardcoded value such as `0px`, to get type, we have to create Token object like the following.
```
let token =
    Token::Dimension {
        has_sign: false, value: 0.0, int_value: Some(0), unit: CowRcStr::from("px")
    };
let token_type = token.serialization_type();
```
I am happy if cssparser has a function to get `TokenSerializationType` without creating object like `TokenSerializationType::nothing()`.

zrhoffman:
Fixes #266

Making TokenSerializationType public is useful to describe a CSS string that is already known to be valid in cases when the TokenSerializationTypes of the start and end of the CSS string is needed, but Tokens are not.

This helps with creating a custom_properties::VariableValue from the CSS string of a computed value for CSS Properties and Values, see Mozilla bug 1858305 [1].

[1]: https://bugzilla.mozilla.org/show_bug.cgi?id=1858305

### DIFF
--- a/src/serializer.rs
+++ b/src/serializer.rs
@@ -378,9 +378,10 @@ impl_tocss_for_float!(f32);
 impl_tocss_for_float!(f64);
 
 /// A category of token. See the `needs_separator_when_before` method.
-#[derive(Copy, Clone, Eq, PartialEq, Debug)]
+#[derive(Copy, Clone, Eq, PartialEq, Debug, Default)]
 pub enum TokenSerializationType {
     /// No token serialization type.
+    #[default]
     Nothing,
 
     /// The [`<whitespace-token>`](https://drafts.csswg.org/css-syntax/#whitespace-token-diagram)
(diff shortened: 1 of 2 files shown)


SimonSapin:
Hi! I haven’t been involved in the maintenance of this crate for a few years now.

zrhoffman:
> Hi! I haven’t been involved in the maintenance of this crate for a few years now.

My bad.

r? @emilio :)

emilio (approved):

PR #371, opened 2024-03-20
Rename `master` branch to `main`

mrobinson:
This is in line with the policy decided in the TSC.


mrobinson (commented):
@emilio do you know if there are any parts of Gecko that need to be updated after this change?

emilio (approved):
Nope, looks good!

mrobinson:
@emilio Thanks for the review!

Issue #376, opened 2024-03-31
Add `serde` optional feature for cssparser-color


Loirooriol:
#363 split the color parsing into a separate crate, which has things like https://github.com/servo/rust-cssparser/blob/c3314d15b281f376538cdc91bbb9ce21d5052231/color/lib.rs#L19

However, its `Cargo.toml` doesn't define any `serde` feature.

@emilio

chipnertkj:
@Loirooriol Hi there, coincidentally I needed the same thing done a while ago, here's my PR: https://github.com/servo/rust-cssparser/pull/377

emilio:
Fixed by #388, thanks @chipnertkj

PR #377, opened 2024-04-05
Refactoring, better ability to use `cssparser-color` as a separate crate.

chipnertkj:
- Use case for better `cssparser-color` separation:
Access to private implementation items for parsers and other software using the crate.
- Example use case and motivation:
My own project. I use the crate to serialize and deserialize CSS3-compatible color data from user input and config files. This was impossible without these changes.
- Serde impls for `color` crate items are gated by a new `serde` feature in the inner crate.
- A few public functions were adjusted with new names, as suggested by `clippy` - caused a semantic conflict with std's `Iterator`.
- General refactoring. 

All checks passed on my fork, but I'm not familiar with the guidelines for commits for this repository. It's my first one here. I hope it's up to expectations. 😁
If something's up, be sure to comment.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "cssparser"
 version = "0.33.0"
-authors = [ "Simon Sapin <simon.sapin@exyr.org>" ]
+authors = ["Simon Sapin <simon.sapin@exyr.org>"]
 
 description = "Rust implementation of CSS Syntax Level 3"
 documentation = "https://docs.rs/cssparser/"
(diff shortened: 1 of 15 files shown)


chipnertkj:
Apparently part of the changes were also requested here: https://github.com/servo/rust-cssparser/issues/376

emilio (changes requested):
A lot of these look good but can we avoid the breaking changes in cssparser? Seem unnecessary to me unless I'm missing something.

chipnertkj:
@emilio That's absolutely doable, it's just that I see valid reasoning in the `clippy` hint, enough of it to increase the minor version, especially since the project isn't in a mature state yet. Realistically, would that change be problematic? Genuine question.

emilio:
Yeah, the tokenizer uses the `try` prefix (well, now `try_parse`) for stuff that rewinds, so `try_next` doesn't feel necessarily like a great name. Also, in a sense the tokenizer / parser's `next()` is basically an `Iterator` (over the tokens), even if it returns an extra info for the errors.

It'd also mean we'd have to rewrite [quite a few callers](https://searchfox.org/mozilla-central/search?q=symbol:S_rs_parser%2FParser%23next().&redirect=false). That's not a huge deal if the improvement is clear, but for me at least `try_next` is just a longer and more confusing name...

In any case, those changes should probably be their own PR. I'd prefer to split the different high-level changes in different PRs if that's possible, so that they can be reviewed / merged independently.

chipnertkj:
I see, sure thing! I'll revert the naming and version changes as soon as possible.

chipnertkj:
Okay, that should be it.

emilio (approved):
Thanks, this looks great, but the commit history is a bit of a mess. I sent https://github.com/servo/rust-cssparser/pull/378 with it cleaned-up.

PR #378, opened 2024-04-06
Serde + clippy fixes

emilio:
This is #377 but rebased, without the merge commits, and commits squashed by topic.


chipnertkj:
Much cleaner, glad I could help! 😊

PR #387, opened 2024-04-23
parser: Implement recent syntax spec change to avoid parsing custom properties that look like selectors as rules.

emilio:
This implements the:

> If the first two non-<whitespace-token> values of rule’s prelude are
> an <ident-token> whose value starts with "--" followed by a
> <colon-token>, then...

From https://drafts.csswg.org/css-syntax/#consume-qualified-rule

Mozilla Bug: 1884879
Differential Revision: https://phabricator.services.mozilla.com/D207796


emilio:
Reviewed upstream by @zrhoffman (this incorporates their suggested changes)

PR #388, opened 2024-05-25
Add support for * prefix in declaration names

hamidrezakp:
Hey
This is a hacky fix for adding support of legacy `*` prefixes to `cssparser`.

Currently this fails to be parsed:
```css
p {
    *color: red;
}
```

And there is a good reason to it, This legacy syntax is only used for ie 6/7 and it's not valid css3 grammar too.

But some libraries and old codebases still depend on this kind of properties to work on old browsers.

I know this is not a correct way to do it, but if you have plans to add this feature to parser, i can prepare another fix that uses feature-gates in cargo.toml and has proper parsing methods for handling this kind of syntax.

### DIFF
--- a/src/serializer.rs
+++ b/src/serializer.rs
@@ -211,6 +211,10 @@ where
             hex_escape(digit, dest)?;
             value = &value[1..];
         }
+        if b'*' == value.as_bytes()[0] {
+            dest.write_str("*")?;
+            value = &value[1..];
+        }
         serialize_name(value, dest)
     }
 }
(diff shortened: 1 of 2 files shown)


src/serializer.rs
```diff
@@ -211,6 +211,10 @@ where
             hex_escape(digit, dest)?;
             value = &value[1..];
         }
+        if b'*' == value.as_bytes()[0] {
```
emilio, line 214: This is not correct, this is not an identifier.

emilio (changes requested):
At the very least this would need tests, but not opposed to this in principle if it's not too annoying...

emilio:
I don't think we're particularly interested in it, if some tooling needs it we could accept it I guess but...

That said, I wonder if this could be implemented on top of cssparser too without changes (basically, on the error code path, try to consume the asterisk + ident + colon)

emilio:
Closing for now, given the discussion above, thanks!

PR #389, opened 2024-05-31
Make create no_std compatible

olanod:
Make crate no_std while keeping it backwards compatible with a default `std` feature enabled(only used to impl Error for ParseError).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,10 +32,12 @@ inherits = "release"
 debug = true
 
 [features]
+default = ["std"]
 bench = []
 dummy_match_byte = []
 # Useful for skipping tests when execution is slow, e.g., under miri
 skip_long_tests = []
+std = []
 
 [workspace]
 members = [".", "./macros", "./color"]
--- a/color/lib.rs
+++ b/color/lib.rs
@@ -1,7 +1,7 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
-
+#![cfg_attr(not(test), no_std)]
 #![deny(missing_docs)]
 
 //! Fairly complete css-color implementation.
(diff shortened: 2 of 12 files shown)


LastLeaf:
It seems that the changes are not complete. I still find some deps need changes or `default-features = false` .

olanod:
I tested it with `cargo build --no-default-features` and `cargo test` with default features on purpose to use std. Is that what you tried?

LastLeaf:
> I tested it with `cargo build --no-default-features` and `cargo test` with default features on purpose to use std. Is that what you tried?

I built with `--target armv7a-none-eabi` (which does not have a std) and fixed some of the compilation errors. You can take a look at [my branch](https://github.com/LastLeaf/rust-cssparser/tree/no-std-support). It also requires a [dep change](https://github.com/upsuper/dtoa-short/pull/9).

olanod:
Merged your branch as your PR for the dependency was merged and released.

src/math.rs
```diff
@@ -0,0 +1,21 @@
+
```
emilio, line 1: Please add a license header.

src/math.rs
```diff
@@ -0,0 +1,21 @@
...
+}
+
+pub(crate) fn f32_round(val: f32) -> f32 {
+    #[cfg(feature = "std")]
+    { val.round() }
+    #[cfg(not(feature = "std"))]
+    { libm::roundf(val) }
+}
+
+pub(crate) fn f64_pow(a: f64, b: f64) -> f64 {
```
emilio, line 16: Probably `#[inline]` these?

emilio (approved):
looks ok, but can you rebase and add the license headers? Thanks

color/Cargo.toml
```diff
@@ -12,10 +12,13 @@ edition = "2021"
 path = "lib.rs"
 
 [dependencies]
-cssparser = { path = ".." }
-serde = { version = "1.0", features = ["derive"], optional = true }
+cssparser = { path = "..", default-features = false }
+serde = { version = "1.0", default-features = false, features = ["derive"], optional = true }
+libm = "0.2.8"
```
emilio, line 17: Ah, I missed this. It seems this dependency should be optional, and only enabled if `std` is disabled?

Cargo.toml
```diff
@@ -21,21 +21,24 @@ encoding_rs = "0.8"
 
 [dependencies]
 cssparser-macros = { path = "./macros", version = "0.6.1" }
-dtoa-short = "0.3"
+dtoa-short = "0.3.5"
```
emilio, line 24: Is this change needed?

src/math.rs
```diff
@@ -0,0 +1,27 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#[inline]
+pub(crate) fn f32_trunc(val: f32) -> f32 {
+    #[cfg(feature = "std")]
+    { val.round() }
+    #[cfg(not(feature = "std"))]
+    { libm::roundf(val) }
+}
```
nicoburns, line 11: This seems to be a typo. Should call into trunc function not round functions

PR #395, opened 2024-11-07
Bump cssparser_color version.

emilio:


nicoburns (changes requested):
I think this needs to be a `0.2.0` release because the `cssparser` version bump was a semver incompatible version bump (as far as cargo is concerned), and there are `cssparser` types in the public API of `cssparser_color`.

emilio:
> I think this needs to be a `0.2.0` release because the `cssparser` version bump was a semver incompatible version bump (as far as cargo is concerned), and there are `cssparser` types in the public API of `cssparser_color`.

Nice catch, thanks!

nicoburns (approved):

PR #398, opened 2025-02-09
Fix handling of backslashes and Unicode escape sequences in CSS content

liulinboyi:
This PR addresses an issue in the handling of backslashes and Unicode escape sequences in CSS content. The current implementation may incorrectly escape backslashes, which can break valid CSS syntax, especially when dealing with Unicode escape sequences like \00a0.

### Changes:
Improved Handling of Backslashes:
The code now correctly distinguishes between backslashes used for escaping specific characters (like " or \) and those that are part of Unicode escape sequences.
If a backslash is followed by a valid Unicode escape sequence (e.g., \00a0), it is preserved as-is.
If a backslash is followed by any other character, it is properly escaped to \\.
Added Validation for Unicode Escape Sequences:
A helper function is_valid_unicode_escape is introduced to check if a sequence of characters following a backslash forms a valid Unicode escape sequence.
This ensures that only valid sequences are preserved, while invalid sequences are treated as regular text.
Code Changes:
```rust
fn handle_backslash(s: &str, i: usize) -> Option<&'static str> {
    if i + 1 < s.len() {
        match s.as_bytes()[i + 1] {
            b'0'..=b'9' | b'a'..=b'f' | b'A'..=b'F' => {
                // If the character following the backslash is part of a Unicode escape sequence, preserve the entire sequence
                let mut j = i + 1;
                while j < s.len() && s.as_bytes()[j].is_ascii_hexdigit() && j - i < 6 {
                    j += 1;
                }
                if j - i > 1 {
                    // Preserve the entire Unicode escape sequence
                    Some("\\")
                } else {
                    // If it is not a valid Unicode escape sequence, escape the backslash itself
                    Some("\\\\")
                }
            }
            _ => {
                // If the character following the backslash is any other character, escape the backslash itself
                Some("\\\\")
            }
        }
    } else {
        // If the backslash is the last character, escape the backslash itself
        Some("\\\\")
    }
}
```

### DIFF
--- a/src/serializer.rs
+++ b/src/serializer.rs
@@ -311,6 +311,36 @@ where
     }
 }
 
+fn handle_backslash(s: &str, i: usize) -> Option<&'static str> {
+    let bytes = s.as_bytes();
+    if i + 1 < s.len() {
+        match bytes[i + 1] {
+            b'0'..=b'9' | b'a'..=b'f' | b'A'..=b'F' => {
+                // If the character following the backslash is part of a Unicode escape sequence, preserve the entire sequence
+                let mut j = i + 1;
+                while j < s.len() && bytes[j].is_ascii_hexdigit() && j - i < 6 {
+                    j += 1;
+                }
+                if j - i > 1 {
+                    // Preserve the entire Unicode escape sequence
+                    Some("\\")
+                } else {
+                    // If it is not a valid Unicode escape sequence, escape the backslash itself
+                    Some("\\\\")
+                }
+            }
+            _ => {
+                // If the character following the backslash is any other character, escape the backslash itself
+                Some("\\\\")
+            }
+        }
+    } else {
+        // If the backslash is the last character, escape the backslash itself
+        Some("\\\\")
+    }
+}
+
+
 impl<W> fmt::Write for CssStringWriter<'_, W>
 where
     W: fmt::Write,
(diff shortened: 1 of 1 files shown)


emilio (changes requested):
> This PR addresses an issue in the handling of backslashes and Unicode escape sequences in CSS content. The current implementation may incorrectly escape backslashes, which can break valid CSS syntax, especially when dealing with Unicode escape sequences like \00a0.

Do you have an example of the broken serialization? This should have a test at least. But I don't understand off-hand how this is correct.

emilio:
In particular, you're changing the serialization of an already unescaped CSS string. So `\00a0` shouldn't show up there as the exact literals, but as the actual codepoint (and thus shouldn't hit the code-path).

emilio:
Ok, pretty sure this is incorrect. Here's a test:

```rust
diff --git a/src/tests.rs b/src/tests.rs
index 3c122f0..ca90faa 100644
--- a/src/tests.rs
+++ b/src/tests.rs
@@ -1351,3 +1351,20 @@ fn servo_define_css_keyword_enum() {

     assert_eq!(UserZoom::from_ident("fixed"), Ok(UserZoom::Fixed));
 }
+
+#[test]
+fn serialize_backslash_roundtrips() {
+    fn parse_string(css_string: &str) -> String {
+        let mut input = ParserInput::new(&css_string);
+        let mut input = Parser::new(&mut input);
+        input.expect_string().unwrap().as_ref().into()
+    }
+    let parsed_string = parse_string(r#""\00a0""#);
+    assert_eq!(parsed_string, "\u{a0}");
+
+    // Serialization round-trips.
+    let mut serialized_string = String::new();
+    super::serializer::serialize_string(&parsed_string, &mut serialized_string).unwrap();
+    assert_eq!(parse_string(&serialized_string), parsed_string);
+}
```

That test passes, and the serialization code generates a compatible string. It seems the issue is in the caller trying to serialize something that has been already escaped... or something along those lines.

PR #406, opened 2025-08-06
chore(deps): update phf to 0.12

xu-cheng:


mrobinson:
@emilio Are you okay with this upgrade? If so, I'll go ahead and merge it.

PR #409, opened 2025-09-28
Fix is_ident_start signature to take immutable reference

nnabeyang:
This PR changes the signature of is_ident_start to accept &Tokenizer instead of &mut Tokenizer, since it does not need mutable access.


emilio (approved):
Thanks!

PR #425, opened 2026-03-09
Make all proc macros optional

nicoburns:
This PR makes all proc macros optional behind feature flags (and removes one completely, replacing it with const evaluation):

- Replace the `dummy_match_byte` feature (which **disables** the optimised `match_byte!` macro and replaces it with a simple `match` statement) with a `fast_match_byte` feature which **enables** the optimised `match_byte!` macro. This feature is enabled by default, but means if you build with `--no-default-features` then you don't get the proc macro.
- Replace the `_cssparser_internal_max_len` proc macro with const evaluation.
- Building on the above two changes, make the `cssparser-macros` dep optional (only enabled by `fast_match_byte` feature)
- ~~Improve the syntax of the `ascii_case_insensitive_phf_map!` macro (see below)~~ removed
- Make `phf` optional for color keyword lookup. This is gated behind a `fast_match_color` feature flag. When the flag is disabled the 

All of the new feature flags are enabled by default, so if you use `cssparser` with default features enabled there is no change except for the `ascii_case_insensitive_phf_map!` macro syntax.

**I have also bumped MSRV from 1.68 to 1.71. This is not actually related to this PR, but is required by recent patch versions of `syn` (in fact, with this PR `cssparser` builds down to 1.63 with default features disabled, but I figure 1.71 is probably old enough?)**

With this PR dependency tree with no default features is just:

```
cssparser v0.35.0 (/Users/nico/code/oss/cssparser)
├── dtoa-short v0.3.3
│   └── dtoa v0.4.8
├── itoa v1.0.6
└── smallvec v1.10.0
```

With default features it is:

```
cssparser v0.35.0 (/Users/nico/code/oss/cssparser)
├── cssparser-macros v0.6.1 (proc-macro) (/Users/nico/code/oss/cssparser/macros)
│   ├── quote v1.0.38
│   │   └── proc-macro2 v1.0.101
│   │       └── unicode-ident v1.0.8
│   └── syn v2.0.98
│       ├── proc-macro2 v1.0.101 (*)
│       ├── quote v1.0.38 (*)
│       └── unicode-ident v1.0.8
├── dtoa-short v0.3.3
│   └── dtoa v0.4.8
├── itoa v1.0.6
├── phf v0.13.1
│   ├── phf_macros v0.13.1 (proc-macro)
│   │   ├── phf_generator v0.13.1
│   │   │   ├── fastrand v2.3.0
│   │   │   └── phf_shared v0.13.1
│   │   │       └── siphasher v1.0.1
│   │   ├── phf_shared v0.13.1 (*)
│   │   ├── proc-macro2 v1.0.101 (*)
│   │   ├── quote v1.0.38 (*)
│   │   └── syn v2.0.98 (*)
│   └── phf_shared v0.13.1 (*)
└── smallvec v1.10.0
```

<details><summary><del>`ascii_case_insensitive_phf_map` syntax change</del> removed</summary>

### `ascii_case_insensitive_phf_map` syntax

The macro syntax was previously:

```rust
cssparser::ascii_case_insensitive_phf_map! {
    keywords -> (u8, u8, u8) = {
        "red" => (255, 0, 0),
        "green" => (0, 255, 0),
        "blue" => (0, 0, 255),
    }
}
```

```rust
cssparser::ascii_case_insensitive_phf_map! {
    static keywords : (u8, u8, u8) = {
        "red" => (255, 0, 0),
        "green" => (0, 255, 0),
        "blue" => (0, 0, 255),
    }
}
```

This uses the conventional Rust `:` syntax for defining a type. And also makes it clearer that the `keywords` bit is an `ident` that is being defined.

</details>

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -21,13 +21,11 @@ jobs:
           - 1.68.0
         features:
           -
-          - --features dummy_match_byte
+          - --no-default-features
           - --features malloc_size_of
         include:
           - toolchain: nightly
             features: --features bench
-          - toolchain: nightly
-            features: --features bench,dummy_match_byte
     steps:
       - uses: actions/checkout@v2
 
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -18,7 +18,7 @@ jobs:
           - nightly
           - beta
           - stable
-          - 1.68.0
+          - 1.71.0
         features:
           -
           - --no-default-features
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,7 +10,7 @@ readme = "README.md"
 keywords = ["css", "syntax", "parser"]
 license = "MPL-2.0"
 edition = "2018"
-rust-version = "1.68"
+rust-version = "1.71"
 
 exclude = ["src/css-parsing-tests/**", "src/big-data-url.css"]
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,7 @@ encoding_rs = "0.8"
 cssparser-macros = { path = "./macros", version = "0.6.1", optional = true }
 dtoa-short = "0.3"
 itoa = "1.0"
-phf = { version = "0.13.1", features = ["macros"] }
+phf = { version = "0.13.1", features = ["macros"], optional = true }
 serde = { version = "1.0", features = ["derive"], optional = true }
 malloc_size_of = { version = "0.1", default-features = false, optional = true }
 smallvec = "1.0"
@@ -33,9 +33,10 @@ inherits = "release"
 debug = true
 
 [features]
-default = ["fast_match_byte"]
+default = ["fast_match_byte", "fast_match_color"]
 bench = []
 fast_match_byte = ["dep:cssparser-macros"]
+fast_match_color = ["dep:phf"]
 # Useful for skipping tests when execution is slow, e.g., under miri
 skip_long_tests = []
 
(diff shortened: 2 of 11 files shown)


emilio (changes requested):
Could you split the less controversial patches into their own PR?

This patch breaks selectors and stylo pretty massively, btw... All of it fixable, but some of the fixes that are needed are a bit annoying, e.g.:

```diff
diff --git a/servo/components/style/media_queries/media_query.rs b/servo/components/style/media_queries/media_query.rs
index 9370c0a2a184..f0aad9b52ed4 100644
--- a/servo/components/style/media_queries/media_query.rs
+++ b/servo/components/style/media_queries/media_query.rs
@@ -50,7 +50,11 @@ impl MediaType {
         // Here we also perform the to-ascii-lowercase part of the serialization
         // algorithm: https://drafts.csswg.org/cssom/#serializing-media-queries
         match_ignore_ascii_case! { name,
-            "not" | "or" | "and" | "only" | "layer" => Err(()),
+            "not" => Err(()),
+            "or" => Err(()),
+            "and" => Err(()),
+            "only" => Err(()),
+            "layer" => Err(()),
             _ => Ok(MediaType(CustomIdent(Atom::from(string_as_ascii_lowercase(name))))),
         }
     }
```

emilio (changes requested):
Two nits, but looks good with those, thanks.

emilio (approved):
Thanks! Consider squashing the "nits" commits into the relevant ones?

Loirooriol:
> When the flag is disabled the

Text got cut

PR #427, opened 2026-03-15
Make color and macros crates default members

nicoburns:
This means that running unqualified `cargo check/build/test/clippy` will run against all 3 crates by default, making it harder to accidentally miss build issues in `cssparser-color` crate which previously wouldn't be checked by default.


emilio:
CI is very unhappy, but looks reasonable with that fixed.

emilio (approved):

PR #428, opened 2026-03-16
Release v0.37.0

nicoburns:
A release that includes https://github.com/servo/rust-cssparser/pull/425 (I'm wanting to use this for value parsing in Taffy, which is currently very low dependency).

Note that `cssparser-color` v0.4 was never published (see https://github.com/servo/rust-cssparser/issues/420). And I think it would still make sense to do so. I believe that all that should be required is to checkout commit 9339ddd71443463dfb85d1185ad50cd98b34bc8f which did the version bump for `cssparser` 0.36 and `cssparser-color` 0.4, and run `cargo publish -p cssparser-color`.

It may make sense to align the version numbers of the `-color` and `-macros` crates with the main `cssparser` crate, but I have not done so in this PR.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "cssparser"
-version = "0.36.0"
+version = "0.37.0"
 authors = ["Simon Sapin <simon.sapin@exyr.org>"]
 
 description = "Rust implementation of CSS Syntax Level 3"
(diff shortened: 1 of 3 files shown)


emilio (approved):

nicoburns:
@emilio When you get a chance, could you hit the publish button on this? (+ ideally `cssparser-color` v0.4 - see PR description above)

emilio:
Both things are done.

nicoburns:
Thank you!

PR #433, opened 2026-07-22
Upgrade `phf` to 0.14

nicoburns:
There is potentially a decent speed boost available behind the experimental new `ptrhash` feature (see: https://github.com/rust-phf/rust-phf/pull/392). But that is not enabled here.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,7 @@ smallvec = "1.0"
 # Optional dependencies
 cssparser-macros = { path = "./macros", version = "0.7.0", optional = true }
 malloc_size_of = { version = "0.1", default-features = false, optional = true }
-phf = { version = "0.13.1", features = ["macros"], optional = true }
+phf = { version = "0.14.0", features = ["macros"], optional = true }
 serde = { version = "1.0", features = ["derive"], optional = true }
 
 [dev-dependencies]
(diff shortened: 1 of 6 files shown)


mrobinson:
```
Run cargo fmt --check && cargo clippy -- -Dwarnings
    Updating crates.io index
 Downloading crates ...
  Downloaded phf_macros v0.14.0
error: failed to download replaced source registry `crates-io`

Caused by:
  failed to parse manifest at `/home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/phf_macros-0.14.0/Cargo.toml`

Caused by:
  failed to parse the `edition` key

Caused by:
  this version of Cargo is older than the `2024` edition, and only supports `2015`, `2018`, and `2021` editions.
Error: Process completed with exit code 101.
```

emilio:
Bumping the msrv might be acceptable.

emilio (approved):

Cargo.toml
```diff
@@ -22,7 +22,7 @@ smallvec = "1.0"
 # Optional dependencies
 cssparser-macros = { path = "./macros", version = "0.7.0", optional = true }
 malloc_size_of = { version = "0.1", default-features = false, optional = true }
-phf = { version = "0.13.1", features = ["macros"], optional = true }
+phf = { version = "0.14.0", features = ["macros"], optional = true }
```
emilio, line 25: Also if this now doesn't require syn 2.0, let's just update to syn 3.0 exclusively while at it.

SimonSapin:
> Bumping the msrv might be acceptable.

I’ve pushed to the PR to do so

Issue #434, opened 2026-07-25
needs_separator_when_before omits CDC, so serialized tokens re-parse wrong


hey-jj:
`TokenSerializationType::needs_separator_when_before` returns false when a `Number`, `DelimHash`, `DelimAt`, or `DelimMinus` is followed by `CDC`. Concatenating the two serializations produces text that tokenizes into different tokens than the input.

## Reproducer

cssparser 0.37.0, default features.

```rust
use cssparser::{Parser, ParserInput, ToCss};

fn main() {
    let mut pi = ParserInput::new("5 -->");
    let mut p = Parser::new(&mut pi);
    let a = p.next().unwrap().clone(); // Number { value: 5.0, int_value: Some(5) }
    let b = p.next().unwrap().clone(); // CDC

    // no separator requested
    assert!(!a
        .serialization_type()
        .needs_separator_when_before(b.serialization_type()));

    let mut s = String::new();
    a.to_css(&mut s).unwrap();
    b.to_css(&mut s).unwrap();
    assert_eq!(s, "5-->");
}
```

## Observed vs expected

Observed: the call returns false and serialization yields `5-->`. Re-parsing `5-->` gives `Dimension { value: 5.0, unit: "--" }` followed by `Delim('>')`.

Expected: true, which signals that an empty comment is needed so the output round-trips. The doc comment on `needs_separator_when_before` says it returns true if "an empty comment /**/ needs to be inserted between them so that they are not re-parsed as a single token". CSS Syntax Level 3 §Serialization requires the serialized form to round-trip.

## Root cause

src/serializer.rs:519. The `Ident` and `AtKeywordOrHash | Dimension` rows list `CDC`. The rows covering `Number`, `DelimHash`, `DelimAt`, and `DelimMinus` do not.

## Scope

Four pairs return false: `Number` then `CDC` gives `5-->`, re-parsing as `Dimension{5,"--"}` + `Delim('>')`. `DelimHash` then `CDC` gives `#-->`, re-parsing as `IDHash("--")` + `Delim('>')`. `DelimAt` then `CDC` gives `@-->`, re-parsing as `AtKeyword("--")` + `Delim('>')`. `DelimMinus` then `CDC` gives `--->`, re-parsing as `Ident("---")` + `Delim('>')`.

Any consumer using this API to re-emit CSS is affected. Stylo calls it for custom-property and `var()` substitution serialization. Present on main as of the 2026-07-21 commit.

PR #435, opened 2026-07-27
serializer: Fix separator decisions for comments and CDC

binggao1230:
With comments preserved, `/*a*/*` never converges: it re-parses as 2 tokens, then 3, then 4, then 5, gaining a `Comment("")` every pass (I ran it to 13). `Token::Comment` is categorised as `DelimSlash`, so a comment inherits the rule for a bare `/` and gets a `/**/` before `*` — but a comment is already closed by its own `*/`, and with comments preserved that separator is itself a token, so the next pass repeats. The opposite omission is #434: `needs_separator_when_before` returns false for `Number`, `#`, `@` and `-` before a CDC, so `5` and `-->` serialise to `5-->` and re-parse as `Dimension { 5, "--" }` + `Delim('>')`. [CSS Syntax 3 §9](https://drafts.csswg.org/css-syntax-3/#serialization) marks the CDC column for all four of those rows and has no comment row at all.

I checked the whole table rather than the five entries: the new test walks every ordered pair of 68 sample tokens and asserts that when the table says no separator is needed, the two serialisations really do re-parse to that same pair. It exempts only consecutive whitespace (which the round-trip requirement allows to collapse) and a block opener with its own closer, and it turned up nothing else — three other `true` entries are redundant but harmless (`Number` before `-`; `|` before `|` or `|=`, left over from the `Column` token removed in 7804843), so I left them alone. `serializer_preserving_comments` was skipping the separator call entirely in that mode; with that guard dropped and an idempotence assertion added it now fails on `/*/*///** /* **/*//* ` from the existing corpus.

`cargo test --all` goes 45 → 47, plus `cargo fmt --check`, `cargo clippy -- -Dwarnings`, the three feature sets, 1.71.0 and miri. The comment-preserving half of the fuzz target is re-enabled; the numeric assertion stays off, since `9999995e-45` → `10e-39` is dtoa-short's rounding rather than this crate's.

### DIFF
--- a/fuzz/fuzz_targets/cssparser.rs
+++ b/fuzz/fuzz_targets/cssparser.rs
@@ -80,7 +80,8 @@ fn fuzz(data: &str, preserving_comments: bool) {
     // TODO: This should ideally pass, but it doesn't for numbers near our
     // precision limits, so parsing e.g., 9999995e-45 generates a serialization
     // of 10e-39 because of dtoa rounding, and parsing _that_ generates a
-    // serialization of 1e-38.
+    // serialization of 1e-38. That is the only remaining source of
+    // non-idempotence, and it comes from dtoa-short, not from this crate.
     //
     // assert_eq!(
     //     serialization, reserialization,
@@ -90,7 +91,5 @@ fn fuzz(data: &str, preserving_comments: bool) {
 
 libfuzzer_sys::fuzz_target!(|data: &str| {
     fuzz(data, false);
-    // TODO(emilio): Serialization when preserving comments is not idempotent.
-    // But in browsers we never preserve comments so that's ok...
-    // fuzz(data, true);
+    fuzz(data, true);
 });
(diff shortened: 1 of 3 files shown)


mrobinson:
@gaoflow Was this bug report and fix generated with an LLM?

binggao1230:
Fair question, and the honest answer is yes — I use LLM-based tooling to find and fix these, working under my direction. I'm accountable for what I submit.

Two things worth separating, though. Issue #434 isn't mine — that was filed by `hey-jj`, an unrelated automated filer. I found the same gap independently and credited the issue because it overlaps.

For the fix itself: the separator-table gap came out of differential-testing the serializer against the tokenizer — round-trip every token pair, re-tokenize, and flag any pair whose serialization doesn't re-parse to the same two tokens. Reverting my change to `src/serializer.rs` while keeping the new tests reddens exactly three of them. I also ran MSRV 1.71, miri, `--no-default-features` and `malloc_size_of`.

If LLM-assisted contributions aren't welcome here, say so and I'll close it — no argument, and I won't open others in servo repos.

mrobinson:
Here's the project's LLM policy: https://book.servo.org/contributing/getting-started.html#ai-contributions

PR #437, opened 2026-08-10
parser: Introduce a nested block limit.

emilio:
And make it default to a reasonable value, but still disable-able.


nicoburns (approved):
Seems reasonable.

tiaanl (approved):

PR #442, opened 2026-08-21
Simplify macros and prepare 0.38.0 release

SimonSapin:
Release motivated by https://github.com/servo/rust-cssparser/pull/433

Drive-by cleanup: with MSRV past [inline consts](https://blog.rust-lang.org/2024/06/13/Rust-1.79.0/#inline-const-expressions) we no longer need the `MaybeUninit::<[MaybeUninit<u8>; N]>::uninit().assume_init()` trick.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "cssparser"
-version = "0.37.0"
+version = "0.38.0"
 authors = ["Simon Sapin <simon.sapin@exyr.org>"]
 
 description = "Rust implementation of CSS Syntax Level 3"
--- a/color/Cargo.toml
+++ b/color/Cargo.toml
@@ -13,7 +13,7 @@ rust-version = "1.85"
 path = "lib.rs"
 
 [dependencies]
-cssparser = { path = "..", version = "0.37", default-features = false }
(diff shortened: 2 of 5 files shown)


emilio (approved):

Loirooriol:
Any reason 0.38 wasn't published to crates.io? Stylo needs eff8704e994dc54e5df2a936b78b51828ee21464 to import https://phabricator.services.mozilla.com/D317537

SimonSapin:
No reason to delay it if we need it, I just haven’t needed it yet for https://github.com/servo/servo/issues/47431. I few other PRs have been merged since this one, so pick whichever point in the history is more convenient for the Gecko import. Just make sure to push a corresponding git tag (or I can help do that, ping me wherever)

Loirooriol:
I think it would be confusing to publish 0.38 as a later commit. So I think 0.38 should be published here on 30658ed3e9402e39d8ee9c569bd2e723d0719913.

And then Stylo will actually need a 0.39 that includes 84af7967fc0093c81f44c2d03029fe6db546485e

SimonSapin:
I think that’s what git tags are for but if you feel strongly about it go ahead. (Do you have access?)

PR #446, opened 2026-08-27
parser: Remove ParserInput.

emilio:
This avoids unnecessary indirection during parsing. We only create nested parsers in two places, and it seems easy to avoid.

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -220,55 +220,23 @@ impl<E: fmt::Display> fmt::Display for ParseError<E> {
 
 impl<E: fmt::Display + fmt::Debug> std::error::Error for ParseError<E> {}
 
-/// The owned input for a parser.
-pub struct ParserInput<'i> {
+/// A CSS parser that borrows its `&str` input, yields `Token`s, and keeps track of nested blocks
+/// and functions.
+pub struct Parser<'i> {
     tokenizer: Tokenizer<'i>,
     cached_token: CachedToken<'i>,
     current_block_depth: u8,
     nested_block_limit: u8,
+    /// If `Some(_)`, .parse_nested_block() can be called.
+    at_start_of: Option<BlockType>,
+    /// For parsers from `parse_until` or `parse_nested_block`
+    stop_before: Delimiters,
 }
 
 struct CachedToken<'i> {
     token: Token<'i>,
     start_position: SourcePosition,
     end_state: ParserState,
-}
-
-impl<'i> ParserInput<'i> {
-    /// 75 nested blocks seems reasonable enough.
-    const REASONABLE_NESTED_BLOCK_LIMIT: u8 = 75;
-
-    /// Create a new input for a parser.
-    pub fn new(input: &'i str) -> ParserInput<'i> {
-        ParserInput {
-            tokenizer: Tokenizer::new(input),
-            nested_block_limit: Self::REASONABLE_NESTED_BLOCK_LIMIT,
-            current_block_depth: 0,
-            cached_token: CachedToken {
-                token: Token::Semicolon,                    // Anything would do.
-                start_position: SourcePosition(usize::MAX), // No token would match this cache.
-                end_state: ParserState::default(),
-            },
-        }
-    }
-
-    /// Sets a limit for how many nested blocks we're allowed to parse. This is useful to avoid
-    /// running out of stack space. By default, it's set to `REASONABLE_NESTED_BLOCK_LIMIT`, but it
-    /// can be overridden or cleared. A limit of 0 will be equivalent to no limit at all.
-    pub fn set_nested_block_limit(&mut self, limit: u8) {
-        self.nested_block_limit = limit;
-    }
-}
-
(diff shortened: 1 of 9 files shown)


emilio:
@SimonSapin you have more historical context than me, can you think of any reason not to do this?

emilio:
This has been reviewed upstream in https://phabricator.services.mozilla.com/D321882 but will wait a bit before merging (at least a day or two) to get performance numbers (though I think we should do this just for simplicity), and also to wait for Simon's or other folk's feedback :)

src/parser.rs
```diff
@@ -1009,40 +992,34 @@ impl<'i: 't, 't> Parser<'i, 't> {
...
-            stop_before: delimiters,
-        };
-        result = delimited_parser.parse_entirely(parse);
-        if error_behavior == ParseUntilErrorBehavior::Stop && result.is_err() {
-            return result;
-        }
-        if let Some(block_type) = delimited_parser.at_start_of {
-            consume_until_end_of_block(block_type, &mut delimited_parser.input.tokenizer);
-        }
+    parser.stop_before = delimiters;
```
emilio, line 998: This is the main tricky change really.

src/parser.rs
```diff
@@ -1094,37 +1071,27 @@ where
...
+    let old_stop_before = parser.stop_before;
+    parser.stop_before = match block_type {
         BlockType::CurlyBracket => ClosingDelimiter::CloseCurlyBracket,
         BlockType::SquareBracket => ClosingDelimiter::CloseSquareBracket,
         BlockType::Parenthesis => ClosingDelimiter::CloseParenthesis,
     };
-    let result;
-    // Introduce a new scope to limit duration of nested_parser’s borrow
-    {
-        let mut nested_parser = Parser {
```
emilio, line 1101: And this one.

nicoburns:
IIRC there's some historical context around this in the git log (on my phone, so can't look it up right now). Would be nice if it can be removed. I think it makes the API a little less flexible around input?

Aside from this change, some docs on how to construct and use Parser (and ParserInput if it stays) would be really nice. The lib.rs docs show how to implement a Parser, but not really how to use one.

emilio:
Err, yeah so... `ParserInput` was introduced in 970e1ca6 (by @jdm), but I think that could've probably just reused `Tokenizer` directly like here. If you blame that further you get to 41fa928b which is the last large rewrite, which had a very different API (`Parser` was returned by value from e.g. `parse_nested_block`), which explains this design a lot more.

So I think this is fine nowadays.

Why do you think this makes the API less flexible? I don't see how.

SimonSapin:
I feel there must have been a reason to bother with two lifetime parameters but right now I can’t remember what it is or was. Stylo is probably the biggest consumer of cssparser, so if you’ve ported it to this new API and feel it’s an improvement I trust your judgment Emilio.

jdm:
> I feel there must have been a reason to bother with two lifetime parameters but right now I can’t remember what it is or was. Stylo is probably the biggest consumer of cssparser, so if you’ve ported it to this new API and feel it’s an improvement I trust your judgment Emilio.

Possibly because the code was written before the stabilization of non-lexical lifetimes in the 2018 edition.

emilio:
Well if you borrow the tokenizer from the parser and create nested parsers you need the two lifetimes, one for the tokenizer, one for the input the tokenizer borrows from.

But if you don't, then one suffices :)

Cool so seems there no reason not to do this... Can I get an stamp if so? Even though this has been reviewed upstream I'd rather not use the big red button to merge :)

SimonSapin:
> Well if you borrow the tokenizer from the parser and create nested parsers you need the two lifetimes, one for the tokenizer, one for the input the tokenizer borrows from.

Ah yes, that rings a bell. Do we not create nested parsers anymore? What changed to lead to that?

emilio:
> Do we not create nested parsers anymore? What changed to lead to that?

Right now we do only internally (e.g. see the two comments in [here](https://github.com/servo/rust-cssparser/pull/446#pullrequestreview-5042897271) which are the two main changes in the PR other than removing the lifetime). But we can just reuse the same parser and restore the state.

In 41fa928b7df1acc05b16af5f4ec3edb229274957, the API was `.parse_nested_block().parse_entirely(parse)`. But that changed in 1796776e, and I don't think we've needed this since then, we could've just fixed up the parser state instead, which is what this PR does.

emilio:
(Rebased over #447)

emilio:
@SimonSapin does it look good given the discussion above?

emilio:
I'll take the upstream review, fairly sure this is ok, and both this and the stylo change is kind of a pain to rebase over and over (plus I want to make further measurement / improvements).

SimonSapin:
Sorry I didn’t get around to reviewing the diff. I’m ok with accepting the previous review in the Gecko patch

emilio:
Yeah I don't think you had to review it necessarily, I mostly asked because I was curious at the historical context for this set-up, but given the blame it seems to be mostly an artifact of how the API used to look like.

PR #447, opened 2026-08-28
parser: Remove Option around the cached token.

emilio:
Avoids some branches in hot loops.


mrobinson (approved):

emilio:
Tweaked the loop to work on our MSRV.

PR #449, opened 2026-08-28
tokenizer: Make consume_name faster.

emilio:
Skip over runs of ascii and such in a tight loop. This is one of the hottest functions and match_byte! adds extra branches for the common cases.

### DIFF
--- a/src/tokenizer.rs
+++ b/src/tokenizer.rs
@@ -312,6 +312,16 @@ impl<'a> Tokenizer<'a> {
         unsafe { self.input.get_unchecked(range.start.0..range.end.0) }
     }
 
+    #[inline]
+    pub(crate) fn byte_slice(&self, range: Range<usize>) -> &'a [u8] {
+        &self.input.as_bytes()[range]
+    }
+
+    #[inline]
+    pub(crate) fn byte_slice_from(&self, start: usize) -> &'a [u8] {
+        self.byte_slice(start..self.position)
+    }
+
     pub fn current_source_line(&self) -> &'a str {
         let current = self.position();
         let start = self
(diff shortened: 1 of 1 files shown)


src/tokenizer.rs
```diff
@@ -917,15 +947,26 @@ fn consume_ident_like<'a>(tokenizer: &mut Tokenizer<'a>) -> Token<'a> {
...
+    // These are the overwhelmingly common bytes, that we can just skip over in a tight loop.
+    static IS_SIMPLE_NAME_BYTE: [bool; 256] = {
+        let mut table = [false; 256];
+        let mut i = 0;
+        while i < 256 {
+            table[i as usize] = matches!(i as u8, b'a'..=b'z' | b'A'..=b'Z' | b'0'..=b'9' | b'_' | b'-' | b'\xC0'..=b'\xEF');
+            i += 1;
+        }
+        table
+    };
```
nicoburns, line 959: Given that this is a table of `bool` you could also try a bitset, and see if that's any faster.

src/tokenizer.rs
```diff
@@ -917,15 +947,26 @@ fn consume_ident_like<'a>(tokenizer: &mut Tokenizer<'a>) -> Token<'a> {
...
+
     // start_pos is the end of the previous token, therefore at a code point boundary
     let start_pos = tokenizer.position();
     let mut value_bytes;
```
nicoburns, line 963: Nit: an explicit type annotation on `value_bytes` would nice.

nicoburns (approved):
This looks correct to me

src/tokenizer.rs
emilio: A bitset was quite a few more instructions: https://rust.godbolt.org/z/G94TTEo7G

src/tokenizer.rs
emilio: Gah I guess the auto-merge got faster. It's not a huge deal tho and it's pre-existing...

PR #453, opened 2026-08-30
chore: Bump rust edition.

emilio:
Our MSRV already supports edition 2024. This came up in #452.

All changes except the tests.rs ones are done by `cargo fix`.

I removed a bunch of replaced expr -> expr_2021 stuff in our macros, because I don't think we care about that distinction.


mrobinson (approved):
Thank you for the followup!

PR #454, opened 2026-09-07
Revert "parser: Remove ParserInput."

mrobinson:
This reverts commit ddccc070a0261cb94195ed0f74095c67b86a8550.
This is a temporary revert in order to release a new version of
`rust-cssparser`.


emilio:
Seems this is not needed after all right?

PR #456, opened 2026-09-15
parser: Collapse parser and tokenizer.

emilio:
Now they are 1:1 so the distinction is mostly internal. This allows
state tracking to be a bit simpler.


emilio:
@SimonSapin not sure how you feel about this, the main point I wanted to get out of this is to be able to not track at_start_of separately, but I think it makes sense in general now that `Parser` and `Tokenizer` are effectively 1:1.

SimonSapin:
Aren’t there tests that run the tokenizer by itsef?

emilio:
> Aren’t there tests that run the tokenizer by itsef?

Not really

emilio:
Fixed CI (msrv didn't support the `if foo && let Some(bar) = ... {` syntax.

src/parser.rs
```diff
@@ -597,41 +504,26 @@ impl<'i> Parser<'i> {
...
 
-        let token_start_position = self.tokenizer.position();
+        let token_start_position = self.position();
         let using_cached_token = self.cached_token.start_position == token_start_position;
-        let token = if using_cached_token {
-            let cached_token = &self.cached_token;
-            self.tokenizer.reset(&cached_token.end_state);
-            if let Token::Function(ref name) = cached_token.token {
-                self.tokenizer.see_function(name)
-            }
```
emilio, line 619: Note that this removal is an intentional change, I want to make the cached token path add as little overhead as possible. This required a change in servo when parsing declaration blocks but I think this is ok. The alternative would be to make `look_for_arbitrary_substitution_functions` invalidate the cached token, at least if it's a function (but we'd need the same change anyway to avoid the performance regression).

emilio (commented):
@SimonSapin does this look reasonable to you? The comment below is the only change that I think is potentially dubious / objectionable in practice, but I'd rather do either nothing (this patch), or the `look_for_arbitrary_substitution_functions` bit suggested below, than the existing thing which penalizes all tokens all the time.

tiaanl:
Tests would be nice, which would also show how using the `Parser` now changes.

emilio:
> Tests would be nice, which would also show how using the `Parser` now changes.

Tests for what? The public API doesn't change at all.

emilio:
FWIW I have a try run with this patch green on Firefox which include all of WPT and mochitests, etc so not too concerned about introducing regressions here. But happy to add coverage for any particular area that you think might deserve it? But seems tangential / worth a separate PR in any case.

tiaanl (approved):
With the caching change handled elsewhere, I don't see issues with it.

SimonSapin:
It’s been years since I’ve properly looked at cssparser code so this isn’t a full review (sorry!) but I’m not opposed to this in principle. Especially having already ported Firefox to this change with green test sounds good. Since that includes Stylo, I assume Servo should be mostly fine too.

emilio:
Yeah pretty sure this change is a no-op except for that very specific `see_function` bit and should be fine for servo. Thanks!

PR #457, opened 2026-09-20
Shrink ParserState

emilio:
See the two commits. This improves state restoration performance which is pretty hot.


emilio:
A less ugly alternative for the last commit might be to just reduce the line count to a `u16`... I'll test that, that gives us some padding for more state later too...

emilio:
Try: https://treeherder.mozilla.org/jobs?repo=try&landoInstance=lando-prod-2025&landoCommitID=94631
SP3: https://perf.compare/compare-lando-results?landoInstance=lando-prod-2025&baseLando=94632&newLando=94633&baseRepo=try&newRepo=try&framework=13
Stylebench: https://perf.compare/compare-lando-results?landoInstance=lando-prod-2025&baseLando=94635&newLando=94636&baseRepo=try&newRepo=try&framework=13

src/parser.rs
```diff
@@ -3,26 +3,43 @@
...
 pub struct ParserState {
     pub(crate) position: usize,
-    pub(crate) current_line_start_position: usize,
-    pub(crate) current_line_number: u32,
-    pub(crate) at_start_of: Option<BlockType>,
+    pub(crate) current_line_start_position: u32,
+    /// Current line number shifted by `Self::BLOCK_TYPE_BITS`, with the low 2 bits holding the
+    /// `BlockType` discriminant of the block the last returned token opened (if any).
+    line_number_and_block_type: u32,
+}
```
emilio, line 23: Further microbenchmarking showed that most of the win (on those microbenchmarks at least) comes from making the `Clone` impl a memcpy rather than a per-field copy (why does rustc not collapse everything onto a memcpy without `derive(Copy)`, I don't know)...

Making it smaller is a smaller win on top, so maybe we should just take this anyway, since I think it's not _too_ complicated...

